Issue #16, opened 2023-10-17
Tests fail because of Typescript issues


Dodecahedr0x:
Running all tests fails because there are issues:
- Some types are invalid
- Some imports don't exist
- Unused code

PR #17, opened 2023-10-17
Fix Typescript issues

Dodecahedr0x:
This PR addresses issue #16,  with the following additions:
- Fix types in test files
- Update to the corresponding version of Anchor JS
- Move unused files to avoid picking them when running all tests


tests/clob.ts
```diff
@@ -11,11 +11,15 @@ describe("clob", () => {
   const provider = anchor.AnchorProvider.env();
   anchor.setProvider(provider);
 
-  const payer = provider.wallet.payer;
+  const payer = anchor.web3.Keypair.generate();
```
metaproph3t, line 14: why introduce a separate payer?

the benefit of the in-built Anchor one is that it will automatically sign transactions, so you don't have to do the `.signers([payer])` biz

tests/conditionalVault.ts
```diff
@@ -907,11 +928,10 @@ async function initializeDepositSlip(
 
 async function generateRandomVault(
   vaultProgram: VaultProgram,
+  payer: Keypair,
+  banksClient: BanksClient,
   settlementAuthority: Keypair = anchor.web3.Keypair.generate()
-): [PublicKey, Keypair, Keypair] {
-  const banksClient = vaultProgram.provider.connection.banksClient;
```
metaproph3t, line 912: did this give a type error or something?

tests/conditionalVault.ts
```diff
@@ -1154,11 +1172,11 @@ export async function redeemDepositSlip(
   assert.equal(
     vaultUnderlyingTokenAccountAfter.amount,
     vaultUnderlyingTokenAccountBefore.amount -
-      BigInt(depositSlipBefore.depositedAmount)
+      BigInt(depositSlipBefore.depositedAmount.toString())
```
metaproph3t, line 1175: it's a bummer that there's not something cleaner here, but yeah it appears that this is the best way to do it on a `BN`

tests/clob.ts
Dodecahedr0x: The problem is that `payer` apparently does not exist in `provider.wallet`. I can however change it to default to Anchor's signer implicitly

tests/conditionalVault.ts
Dodecahedr0x: Yes, there is no `banksClient` in `web3.Connection`

tests/clob.ts
metaproph3t: yeah, that would be good IMO

the way I deal with this in the scripts is `const payer = provider.wallet['payer'];`

metaproph3t:
dw about test failure BTW

that's a me problem

tests/clob.ts
Dodecahedr0x: Looking into it, there is a `payer` only if it's a `NodeWallet`, which `BankrunProvider` provides but not necessarily `anchor.AnchorProvider.env()`. I think that would require switching to Bankrun for `tests/clob.ts` to make it correctly, but that would be tedious in this PR. I can just revert the payer generation for now, and address it in another PR later

metaproph3t: ok, I'll merge it as is, and if it starts feeling non-ergonomic we can always roll back later

Issue #18, opened 2023-10-18
Enable multiple DAO instantiations


Dodecahedr0x:
Currently, the DAO is global for the program. It simplifies some parts of the code but it makes it more expensive to create new DAO instances because you need to redeploy the program. Enabling multiple instances makes it more affordable for everyone to experiment with futarchy. It could also open the door for DAO merging and forking, which might prove helpful.

metaproph3t:
Yeah, I agree that this would be a useful feature

At the moment, I prefer that we keep it as it is in order to minimize the attack / bug surface (e.g., a bug could be a mismatch between a DAO and some vault account), but definitely this could be something worth upgrading to

metaproph3t:
This has been done

PR #19, opened 2023-10-21
Multiple DAO instances

Dodecahedr0x:
This PR addresses issue #18.

Adds an identifier in the DAO account to differentiate between multiple instances.
Adds the necessary  checks to ensure that a proposal is applied to the DAO it was initialized for.

I currently feel like a DAO should be tightly coupled to its CLOB since the DAO treasury could act as the admin, and that this modification should also be applied to the CLOB program.


metaproph3t:
Resolved by other means

PR #24, opened 2023-11-15
Fixes in tests

lagzda:
Updated `getMarket` with `getMarketAccount` and updated the `amount` parameter requirement for `mintConditionalTokens` to support either `number` or `bigint`.


metaproph3t (approved):
LGTM! thanks sir!

PR #36, opened 2023-12-02
Autocrat v0.1

metaproph3t:


programs/autocrat_v0/src/lib.rs
```diff
@@ -381,18 +388,44 @@ pub mod autocrat_v0 {
...
+        Ok(())
+    }
+
+    pub fn set_slots_per_proposal(ctx: Context<Auth>, slots_per_proposal: u64) -> Result<()> {
+        let dao = &mut ctx.accounts.dao;
+
+        dao.slots_per_proposal = slots_per_proposal;
+
+        Ok(())
+    }
```
0xbigz, line 416: it might make more sense to include a single update params function with using `Option`s. that way you could update in conjunction with others more easily

example this: https://github.com/drift-labs/drift-competitions/blob/890830d0e76846f20fcf12036dd495b297bcd032/programs/drift-competitions/src/instructions/update_competition.rs#L51

programs/autocrat_v0/src/lib.rs
```diff
@@ -381,18 +388,44 @@ pub mod autocrat_v0 {
...
+
+    pub fn set_market_taker_fee(ctx: Context<Auth>, market_taker_fee: i64) -> Result<()> {
+        let dao = &mut ctx.accounts.dao;
+
+        dao.market_taker_fee = market_taker_fee;
+
+        Ok(())
```
0xbigz, line 423: would the fees accumulate in the conditional token markets? who is the owner to withdraw form there? the proposal initializer or the dao? if its the dao, would this program be able to if immutable?

0xbigz:
additionally, i think it would be nice if we tried to hammer out (or at least create ways to update/tackle later through proposal) as many autocrat related ones

the scope here is entirely the autocrat program? and many of these are related to vaults

programs/autocrat_v0/src/lib.rs
R-K-H: It needs to be paid out to the makers by makerVolume as a % of total collected fees. 

Not sure if that could be called on market close or how resolution occurs.

Maybe it could be part of the redeem function as it'd be associated with the account?

State gets a bit muddy if people can still trade the markets after proposal is completed, how would you flag an account as having redeemed, but then able to redeem more if the market still trades?

metaproph3t: Right now it just goes to the treasury, we could add the other functionality as a future version of autocrat

metaproph3t:
Proposal passed [here](https://solscan.io/tx/43iJ7sUEUY7BH8qC1najoWjKPWqfxGpYQVEczmUstVbFYgrFt1hQcCakjcFCvjV3geC3zqR6tAhBUw9noKT6PRPs), now merging

PR #41, opened 2024-02-05
Updated script for submitting proposal and creating OB markets

joebuild:
This updates the proposal submission to:
- reflect minor changes in the autocrat program
- tidy up some code to make creating markets and proposal submission less prone to state rent loss
- uses the latest version of the openbook client, where the main change is letting the user handle instructions/transactions on their own (much better imo)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ target
 node_modules
 test-ledger
 scripts/*.js
+.env
+.idea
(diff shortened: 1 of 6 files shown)


metaproph3t (approved):

metaproph3t:
Changes have been incorporated into mainline

PR #47, opened 2024-02-19
Add clockwork twap script

nickgarfield:


scripts/clockworkTwap.ts
```diff
@@ -0,0 +1,143 @@
...
+                    twapMarket: failMarketTwap,
+                    openbookProgram: OPENBOOK_PROGRAM_ID,
+                })
+                .instruction();
+    
+    const threadId = failMarket.toString().substring(0, 8);
+    const [threadAddress] = clockwork.getThreadPDA(payer.publicKey, threadId);
+    const trigger = {
+        cron: {
+            schedule: "*/10 * * * * * *",
```
joebuild, line 118: would suggest doing once a minute (or once every 15 seconds if that is supported)
everyone is happy to pay the fees

joebuild:
can we have this spit out an address where we can send SOL and monitor as fees are spent?

staccDOTsol:
gm nick

Issue #45, opened 2024-02-18
anchor tests are broken


crispheaney:
`createMarket` doesn't exist on openbook client https://github.com/metaDAOproject/futarchy/blob/main/tests/autocratV0.ts#L1458

johnanthos:
The linked issue was merged and tests are passing so I believe this can be closed, right @crispheaney?

crispheaney:
yup sg

PR #51, opened 2024-02-23
fix autocrat anchor tests

jshiohaha:
Addresses issue https://github.com/metaDAOproject/futarchy/issues/45

Most updates were just updating against differences on the openbook client. Otherwise, I installed a previous version of the `@solana/spl-token` so we didn't have to write getAssociatedTokenAddress ourself or import from another package. Everything else should be the same.

It doesn't prove anything, but here's a screenshot of the end of the passing tests 😅 

Screenshot 2024-02-23 at 16 58 17


Henry-E:
Referenced this PR in #54 , the new import is actually causing issues. Maybe we can revert this change or make it clearer what steps are needed to run it without issue because right now when I just run `yarn install` typescript picks the wrong `spl-token` version (the older one).

Issue #40, opened 2023-12-18
Add pMETA and fMETA metadata


metaproph3t:
Self-explanatory. In the script that initializes proposals, add metadata that allows wallets to see that one is pMETA and one is fMETA

PR #52, opened 2024-02-28
feat: add the ability to upload offchain metadata for conditional mints and link the metadata onchain

jshiohaha:
# Description

Add the ability to create both off-chain metadata as JSON on arweave, and pass the resulting URI into the create on-chain metadata instruction for pass and fail mints for both vaults for a proposal.

Addresses Issue #40

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## What changes are included?

- **Contract**: there is a new instruction that simply calls into the token metadata contract twice, once for each of the pass and fail conditional mints. This could also be collapsed into the `initialize_vault` instruction if people don't like have separate instructions. I just kept them separate at first.
- **Scripts**: I added the new instruction to the `initializeVault` function of the script. So, any script operators should be able to run the script as is without any changes.
- **Tests**: I updated the anchor tests so that we create on-chain metadata for conditional mints. I also added the dumped `mpl_token_metadata.so` binary, which will get picked up by the underlying test validator. You simply have to run `anchor test`.

# How Has This Been Tested?

- [x] I updated anchor tests to verify the new instruction works as expected
- [x] I manually ran the upload image and metadata JSON functions from my computer to verify that assets could be successfully uploaded


metaproph3t:
Gonna merge this, @joebuild hopefully you can port this to AMM code

Henry-E:
Things I would have quibbled about before merging this PR
- Removes dependencies on `anchor.web3` and switches them over to straight `@solana/web3.js`. It's important to just the `anchor.web3` version to avoid any potential version mismatches down the road
- Needs independent testing, as far as I can tell typing for spl-token is broken when i git cloned and tried to run some scripts

PR #61, opened 2024-03-09
fix: consolidate @solana/spl-token package versions

### LINKED ISSUE #54: Weird node based import messes with yarn
https://github.com/metaDAOproject/futarchy/blame/593ae6ad449f9110b10087eb0ceebc86020ee3be/package.json#L16

I don't have a huge understanding of the mess that is npm and yarn but this line messes with any attempts to run scripts using yarn. Yarn goes to the wrong spl-token package, even though there's a node based alias and just ignores the correct spl-token library. 

Practically speaking deleting the package from package.json and/or running npm install worked in terms of getting the script to import the correct version of spl-token but then when it came to either compiling or running the scripts using yarn stuff just started to break again.

jshiohaha:
**Description**

If merged, this PR fixes #54. Although to clarify, I was unable to validate the claim that the dual imports were causing any problems in scripts while testing.

**Context**

In previous PR #52, I added another version of `@solana/spl-token` because I thought the newer version was missing a certain function (`getAssociatedTokenAddress`). It turns out the existing package was fine, I simply needed to re-order some of the function's parameters.

I also made sure to follow the pattern around the rest of the repo with an aliased import as follows

```typescript
import * as token from "@solana/spl-token";
```

**Testing**

Ran `anchor test` to verify all tests still passed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,6 @@
         "@noble/secp256k1": "^2.0.0",
         "@openbook-dex/openbook-v2": "0.1.7",
         "@solana/spl-token": "^0.3.7",
-        "@solana/spl-token-018": "npm:@solana/spl-token@0.1.8",
         "@solana/web3.js": "^1.90.0",
         "@sqds/multisig": "^2.0.0",
         "anchor-bankrun": "^0.2.0",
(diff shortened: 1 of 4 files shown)


R-K-H:
LGTM, however this isn't my domain of PR hell, so I'll leave it to someone who knows the repo better.

metaproph3t:
@Henry-E can you verify that this fixes the issue for you so I can merge PR + close issue?

Henry-E:
Tested this locally and it works. Thanks for the quick fix.

PR #65, opened 2024-03-13
feat: prop-12

R-K-H:
feat: adds prop-12 details, more logging and checks

### DIFF
--- a/scripts/initializeProposal.ts
+++ b/scripts/initializeProposal.ts
@@ -2,6 +2,7 @@ import { initializeProposal, daoTreasury, META } from "./main";
 import * as anchor from "@coral-xyz/anchor";
 import { MEMO_PROGRAM_ID } from "@solana/spl-memo";
 import * as token from "@solana/spl-token";
+import { LAMPORTS_PER_SOL } from "@solana/web3.js";
 
 const { PublicKey, Keypair, SystemProgram } = anchor.web3;
 const { BN, Program } = anchor;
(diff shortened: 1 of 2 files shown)


scripts/initializeProposal.ts
```diff
@@ -11,9 +12,26 @@ anchor.setProvider(provider);
...
+const COST_DEPLOY = 4 * LAMPORTS_PER_SOL
 
 async function main() {
+  console.log('Initializing with PublicKey');
+  console.log(payer.publicKey.toString());
+
+  const payerBalance = await provider.connection.getBalance(payer.publicKey);
+  console.log('Account SOL balance');
+  console.log(payerBalance / LAMPORTS_PER_SOL);
+  if (payerBalance < COST_DEPLOY){
```
LukasDeco, line 26: isn't `COST_DEPLOY` represented as lamports because it's `4*LAMPORTS_PER_SOL`? But I could be totally missing something here. But then it looks like we are comparing SOL to lamports here.

LukasDeco: nope nvm. Still waking up

Henry-E: Not sure what you're getting at. This looks correct. Everything is represented in terms of lamports (9 decimals).

LukasDeco: yeah I was looking at it incorrectly. Hence my updated comment

R-K-H:
I want to include this into the repo given it sets up some pretty good guardrails for people wanting to deploy. The next step would be to ensure we add priority fees everywhere as well as some places where we end up re-creating things we don't need to if the process stalls (looking at openbook markets).

PR #66, opened 2024-03-14
Merge tokens

Henry-E:
Adds the ability for users to merge pass/fail tokens back to their underlying token before the end of the market.

It was lazy but essentially the logic is all copied and pasted and minorly modified from the existing `redeem` function and tests.

Small note: I chose to take an additive approach to the program. But it would have been possible to reuse the account struct `RedeemConditionalTokensForUnderlyingTokens` except that it has a constraint involving the `vault status` which could possibly be inside the `redeem` function as a `require(vaultStatus, Finalized)` check. Rather than changing the existing code though, I just copy/pasted to a new account struct and modified the constraint.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ and what they do can be found at [docs.themetadao.org](https://docs.themetadao.o
 ## Scripts
 
 The scripts folder contains a few scripts that you can use to interact with the Meta-DAO.
-Today, the only way to create proposals is via script. You can do this by modifying the 
+Today, the only way to create proposals is via script. You can do this by modifying the
 `initializeProposal.ts` script and replacing its `pubkey`, `accounts`, and `data` with the
 SVM instruction that you want to use in your proposal.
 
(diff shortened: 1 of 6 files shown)


Henry-E:
One more commit still to come with the program ID updates

Henry-E:
Uploaded to devnet, ready to go for mainnet once CLI stuff gets sorted out

https://solscan.io/account/vAuLTQjV5AZx5f3UgE75wcnkxnQowWxThn1hGjfCVwP?cluster=custom&customUrl=https%3A%2F%2Fapi.devnet.solana.com
```
solana-verify get-program-hash -u $NETWORK_URL $PROGRAM_ID
9f3f4b4e56642797f5f296b9be0d424c6cd4244caaf5d10df838ea5f344316cf
```

README.md
```diff
@@ -17,12 +17,13 @@ either devnet, mainnet, or (recommended) an RPC URL.
...
 
-| program           | tag | program ID                                  |
-| ----------------- | --- | ------------------------------------------- |
-| autocrat_v0       | v0.1| metaX99LHn3A7Gr7VAcCfXhpfocvpMpqQ3eyp3PGUUq |
-| autocrat_migrator | v0.1| migkwAXrXFN34voCYQUhFQBXZJjHrWnpEXbSGTqZdB3 |
-| autocrat_v0       | v0  | meta3cxKzFBmWYgCVozmvCQAS3y9b3fGxrG9HkHL7Wi |
-| conditional_vault | v0  | vaU1tVLj8RFk7mNj1BxqgAsMKKaL8UvEUHvU3tdbZPe |
+| program           | tag  | program ID                                  |
+| ----------------- | ---- | ------------------------------------------- |
+| conditional_vault | v0.1 | vAuLTQjV5AZx5f3UgE75wcnkxnQowWxThn1hGjfCVwP |
```
metaproph3t, line 22: this should be v0.2, that's where we're going to pop this

programs/conditional_vault/src/lib.rs
```diff
@@ -155,6 +155,104 @@ pub mod conditional_vault {
...
+
+        let vault = &accs.vault;
+        let vault_status = vault.status;
+        require!(vault_status == VaultStatus::Active, ErrorCode::VaultAlreadySettled);
+
+        // Store Pre-operation Balances
+        let pre_user_conditional_on_finalize_balance = ctx
+        .accounts
+        .user_conditional_on_finalize_token_account
+        .amount;
```
metaproph3t, line 171: weird indentation

programs/conditional_vault/src/lib.rs
```diff
@@ -155,6 +155,104 @@ pub mod conditional_vault {
...
     }
 
+    pub fn merge_conditional_tokens_for_underlying_tokens(
+        ctx: Context<MergeConditionalTokensForUnderlyingTokens>, amount: u64
+    ) -> Result<()> {
+        let accs = &ctx.accounts;
+
+        let vault = &accs.vault;
+        let vault_status = vault.status;
+        require!(vault_status == VaultStatus::Active, ErrorCode::VaultAlreadySettled);
```
metaproph3t, line 165: you already have this covered in the constraints

metaproph3t (approved):
everything LGTM except tag should be 0.2

programs/conditional_vault/src/lib.rs
Henry-E: Yeh, I know it's double checked but I think the check should be happening here instead. I suppose I can just delete it.

programs/conditional_vault/src/lib.rs
Henry-E: It should be auto formatted by cargo fmt but I haven't doubled check that it's doing this correctly for a little while.

README.md
Henry-E: Got it

Issue #67, opened 2024-03-16
Update the migrator program to allow for 4 tokens


metaproph3t:
Today, the [migrator program](https://github.com/metaDAOproject/futarchy/blob/main/programs/autocrat_migrator/src/lib.rs) can only transfer 2 tokens at once. Currently, there are 4 tokens in the treasury that we'd like to migrate:

![Image](https://github.com/metaDAOproject/futarchy/assets/108694867/bf6bd76c-8e61-45c9-8a22-90dd0906e2b5)

We should update the migrator program to allow for this.

metaproph3t:
Completed via #68

PR #68, opened 2024-03-16
Extend migrator to 4

metaproph3t:
Addresses #67

### DIFF
--- a/programs/autocrat_migrator/src/lib.rs
+++ b/programs/autocrat_migrator/src/lib.rs
@@ -2,8 +2,8 @@
 //! to the new autocrat.
 
 use anchor_lang::prelude::*;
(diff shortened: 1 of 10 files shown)


Henry-E:
This looks good. It's unfortunate that it threw in a bunch of linting changes on other programs as well though. I couldn't quite figure out if there were real changes to the tests outside of the migrator test.

Assuming the tests pass locally, there's another easy test we can do once the program is on mainnet. We can construct and simulate the transaction with the requirement for signatures turned off just to verify it works.

Henry-E (approved):
Some extra changes in tests that don't seem related, probably just annoying linting stuff. Left main review comment in the PR itself.

PR #69, opened 2024-03-17
Autocrat Changes

Henry-E:
Some minor changes to autocrat. Mostly config based.

### DIFF
--- a/programs/autocrat_v0/src/lib.rs
+++ b/programs/autocrat_v0/src/lib.rs
@@ -164,7 +164,7 @@ pub mod autocrat_v0 {
             AutocratError::InvalidMarket
         );
         require!(
-            openbook_pass_market.base_lot_size == 1_000_000_000, // minimum tradeable = 1 META
+            openbook_pass_market.base_lot_size == 100_000_000, // minimum tradeable = 0.1 META
             AutocratError::InvalidMarket
         );
         require!(
@@ -203,7 +203,7 @@ pub mod autocrat_v0 {
             AutocratError::InvalidMarket
         );
         require!(
-            openbook_fail_market.base_lot_size == 1_000_000_000,
+            openbook_fail_market.base_lot_size == 100_000_000, // minimum tradeable = 0.1 META
             AutocratError::InvalidMarket
         );
         require!(
--- a/programs/autocrat_v0/src/lib.rs
+++ b/programs/autocrat_v0/src/lib.rs
@@ -29,8 +29,8 @@ declare_id!("metaX99LHn3A7Gr7VAcCfXhpfocvpMpqQ3eyp3PGUUq");
 pub const SLOTS_PER_10_SECS: u64 = 25;
 pub const THREE_DAYS_IN_SLOTS: u64 = 3 * 24 * 60 * 6 * SLOTS_PER_10_SECS;
 
-// by default, the pass price needs to be 5% higher than the fail price
-pub const DEFAULT_PASS_THRESHOLD_BPS: u16 = 500;
+// by default, the pass price needs to be 3% higher than the fail price
+pub const DEFAULT_PASS_THRESHOLD_BPS: u16 = 300;
 
 // start at 10 SOL ($600 at current prices), decay by ~5 SOL per day
 pub const DEFAULT_BASE_BURN_LAMPORTS: u64 = 10 * solana_program::native_token::LAMPORTS_PER_SOL;
--- a/programs/autocrat_v0/src/lib.rs
+++ b/programs/autocrat_v0/src/lib.rs
@@ -38,6 +38,9 @@ pub const DEFAULT_BURN_DECAY_PER_SLOT_LAMPORTS: u64 = 23_150;
 
 pub const MAX_BPS: u16 = 10_000;
 
+// TWAP can only move by $5 per slot
+pub const MAX_OBSERVATION_CHANGE_PER_UPDATE_LOTS: u64 = 50_000;
+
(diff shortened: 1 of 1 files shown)


Henry-E:
Still TODO, getting the tests working again.

programs/autocrat_v0/src/lib.rs
```diff
@@ -29,15 +29,20 @@ declare_id!("metaX99LHn3A7Gr7VAcCfXhpfocvpMpqQ3eyp3PGUUq");
 pub const SLOTS_PER_10_SECS: u64 = 25;
 pub const THREE_DAYS_IN_SLOTS: u64 = 3 * 24 * 60 * 6 * SLOTS_PER_10_SECS;
 
-// by default, the pass price needs to be 5% higher than the fail price
-pub const DEFAULT_PASS_THRESHOLD_BPS: u16 = 500;
+pub const TEN_DAYS_IN_SECONDS: i64 = 10 * 24 * 60 * 60;
+
+// by default, the pass price needs to be 3% higher than the fail price
+pub const DEFAULT_PASS_THRESHOLD_BPS: u16 = 300;
```
Dodecahedr0x, line 35: Why not include that in the DAO config?

Henry-E: I'm not touching more code than I have to really. If it's not already there, I probably won't put it in. At least not right now.

Henry-E:
Thanks for the test fixes @Dodecahedr0x!

programs/autocrat_v0/src/lib.rs
```diff
@@ -145,10 +150,13 @@ pub mod autocrat_v0 {
             AutocratError::InvalidMarket
         );
 
-        // this should also be checked by `openbook_twap`, but why not take the
-        // precaution?
+        let current_time = Clock::get().unwrap().unix_timestamp as i64;
+
+        // The market expires a minimum of 7 days after the end of the proposal.
```
metaproph3t, line 155: 7 or 10?

programs/autocrat_v0/src/lib.rs
```diff
@@ -29,15 +29,20 @@ declare_id!("metaX99LHn3A7Gr7VAcCfXhpfocvpMpqQ3eyp3PGUUq");
...
+pub const DEFAULT_PASS_THRESHOLD_BPS: u16 = 300;
 
 // start at 10 SOL ($600 at current prices), decay by ~5 SOL per day
 pub const DEFAULT_BASE_BURN_LAMPORTS: u64 = 10 * solana_program::native_token::LAMPORTS_PER_SOL;
 pub const DEFAULT_BURN_DECAY_PER_SLOT_LAMPORTS: u64 = 23_150;
 
 pub const MAX_BPS: u16 = 10_000;
 
+// TWAP can only move by $5 per slot
+pub const MAX_OBSERVATION_CHANGE_PER_UPDATE_LOTS: u64 = 50_000;
```
metaproph3t, line 44: actually just realized this should be `5_000` since we've changed the base lot size, also would be good to add a test so we'd catch this next time

programs/autocrat_v0/src/lib.rs
```diff
@@ -227,6 +235,20 @@ pub mod autocrat_v0 {
             openbook_twap_fail_market.twap_oracle.initial_slot + 50 >= clock.slot,
             AutocratError::TWAPMarketTooOld
         );
+        require!(
+            openbook_twap_pass_market
+                .twap_oracle
+                .max_observation_change_per_update_lots
+                == MAX_OBSERVATION_CHANGE_PER_UPDATE_LOTS,
```
metaproph3t, line 242: wdyt about adding these to DAO params?

programs/autocrat_v0/src/lib.rs
Henry-E: 3 days of proposal + 7 days after the end of the proposal

10 days in total minimum.

programs/autocrat_v0/src/lib.rs
Henry-E: Fixed.

programs/autocrat_v0/src/lib.rs
Henry-E: I take it back. `pass threshold bps` is already configurable.

Henry-E:
I'm still learning the tests. Is there a good way to check that the $5 increments is working as intended?

Henry-E:
Anyone else forking this will have to decide on their own values for
- MAX_OBSERVATION_CHANGE_PER_UPDATE_LOTS -> the $ value of the each twap move, in terms of price lots
- expected value -> the starting $ value of the TWAP, in terms of price lots. Meta used to have EV = $1 but moved it up $100 in this change.

They will also need to decide whether the base and quote lot sizes are appropriate. E.g. minimum base lots size of 1 token when the token has 9 decimals, would mean `base_lot_size = 1_000_000_000`. And assuming the quote currency is USDC, which has 6 decimals, then a `quote_lot_size = 100` allows price down to one hundredth of a cent or $0.0001. Meme coins sometimes have even lower prices, so quote lot size might need to be adjusted down further.

Henry-E:
Brief summary of changes
- Set default pass threshold to 3%
- set max observation change per update lots to $5 and make it a configurable option
- set default expected value to $100
- Ensure that the open markets expire a minimum of 10 days from the creation of the proposal to allow for rent retrieval from openbook markets
- Reduce the openbook base lot size so that people can trade in lots of 0.1 meta

PR #78, opened 2024-03-28
Add `ComputeBudgetProgram` instructions to proposal initialize script

metaproph3t:
Increases the transaction land rate for proposal initialization.

### DIFF
--- a/scripts/main.ts
+++ b/scripts/main.ts
@@ -226,6 +226,14 @@ async function initializeVault(
       systemProgram: SystemProgram.programId,
     })
     .signers([conditionalOnFinalizeKP, conditionalOnRevertKP])
+    .preInstructions([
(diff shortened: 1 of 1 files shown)


scripts/main.ts
```diff
@@ -403,13 +411,17 @@ export async function initializeProposal(
...
     microLamports: 100,
   });
+  const cuLimitIx = ComputeBudgetProgram.setComputeUnitLimit({
+    units: 150_000
+  });
 
   let tx1 = new Transaction();
   tx1.add(...passMarketInstructions);
-  tx1.add(cuIx);
+  tx1.add(cuPriceIx);
```
Henry-E, line 423: This looks like the compute budget program instructions get called after the pass market instructions? Does it matter what order the instructions get called in. Only asking because in an early function they're added as preInstructions.

Issue #75, opened 2024-03-26
Expand program to support multi dao (ID)


R-K-H:
Revisions which need to address:
- [ ] Coupling with META token
- [ ] PDA w/ Seed
- [ ] Base unit logics (decimals with tokens)
- [ ] Interface with expansive capacity for iterations (metadata or config settings)

What I'm thinking here is if we want to test things is there a way to securely make changes which afford flexibility without arduous process. I understand this could be in conflict with security so not going to press that issue too much.

Say we'd like to build a proposal with three day see the response, then one day see the response, eventually resolving to a fixed code structure, or creation of structure supporting the results of our testing and evaluation.

For additional context: The question is where should default values of potentially adjustable values live? If you wanted to enforce PreTrading proposal state, but some didn't, or if you wanted to afford 1-10 day proposal time, or if you wanted to adjust minimum order size, or if you wanted to use AMM exclusively compared to hybrid, or only one instruction type.

Those questions exist for me, and the practicality of which are unknown.

metaproph3t:
I've got some time today, going to try and bang this out

PR #80, opened 2024-03-28
Support multi-DAO

metaproph3t:
Allow multiple DAOs to leverage a single program. Addresses #75.


Henry-E:
Are base lot size and twap expected value the only two configurable options that will be allowed when a DAO is initialized? 

What about the quote lot size, it's dumb but you do see a lot of meme coins with very small prices.
To avoid confusion, should expected price be passed in using the absolute value in USDC and then modified to price lots based on the base and quote lot sizes. Can it be trusted DAO creator won't mess this up. Depends on the UI probably.

Also it's taking away the PDA-ness of the DAO account. How much more annoying will this makes for things on the front end.

programs/autocrat_v0/src/lib.rs
```diff
@@ -457,19 +473,13 @@ pub struct InitializeDAO<'info> {
...
     )]
     pub dao: Account<'info, DAO>,
     #[account(mut)]
     pub payer: Signer<'info>,
     pub system_program: Program<'info, System>,
-    #[account(mint::decimals = 9)]
-    pub meta_mint: Account<'info, Mint>,
+    pub token_mint: Account<'info, Mint>,
     #[account(mint::decimals = 6)]
     pub usdc_mint: Account<'info, Mint>,
```
Henry-E, line 484: The pubkey of the USDC mint is known on mainnet and a cfg check could possibly be used to ensure that USDC is used and not another 6 decimals token. Not a massive deal though

metaproph3t:
> Are base lot size and twap expected value the only two configurable options that will be allowed when a DAO is initialized?
> 
> What about the quote lot size, it's dumb but you do see a lot of meme coins with very small prices. To avoid confusion, should expected price be passed in using the absolute value in USDC and then modified to price lots based on the base and quote lot sizes. Can it be trusted DAO creator won't mess this up. Depends on the UI probably.
> 
> Also it's taking away the PDA-ness of the DAO account. How much more annoying will this makes for things on the front end.

Yeah, the UI needs to sort this out: we'd want it all to be configured in the base lot sizes

programs/autocrat_v0/src/lib.rs
metaproph3t: Yep, agreed. Let me create this as an issue.

Henry-E:
The program ID here is `FuTPR6ScKMPHtZFwacq9qrtf9VjscawNEFTb2wSYr1gY` is that going to be the v0.3 devnet and mainnet program ID for the version of autocrat that supports multi-DAO?

PR #82, opened 2024-03-28
Change default proposal count back to 0

metaproph3t:
Now that we have separate production and develop branches, we can do this.


metaproph3t:
We're going to start with 1 instead to appease the normies

PR #97, opened 2024-04-03
Prop 15

metaproph3t:
[Proposal 15](https://app.themetadao.org/proposal?id=15) passed, migrating MetaDAO's assets the the new programs on develop.


Henry-E:
Ah cool, so this is how the develop -> production merging works. One thing to note though, I'm not sure if `solana-verify verify-from-repo` supports verifying against branches, just specific commit hashes. Though maybe it supports branches passively if you can pass an extended length URL directing toward a particular branch, haven't tested it out. It probably works ok if we put everything in develop, put a moratorium on PRs while we deploy and verify, then merge into production.

Issue #72, opened 2024-03-25
Split up finalizing and instruction execution


Henry-E:
If a proposal is successful when it's finalized the instruction stored on the proposal is executed. Instead we should have an instruction that finalizes the proposal and updates the state of the proposal. Then a separate instruction in which successful proposal instructions can be executed. This would avoid some potential locking issues in the finalization instruction.

metaproph3t:
Completed via #98

PR #98, opened 2024-04-03
Separate proposal passing/failing and instruction execution

metaproph3t:
Designed to address #72


Henry-E (commented):
It's already merged but double checking anyway. Looks good.

PR #100, opened 2024-04-04
feat: adds common instructions, checks, verbose

R-K-H:
This improves the script for initializing proposals by:

- including common instructions
- improving feedback to the user
- adding more logging
- checks for balances
- sleep before trigger


R-K-H:
Addressed comments:

1. Renamed functions to better represent their function.
2. Reviewed comment about output of IX, think we should still include it for review. More is better to ensure we catch things.
3. 566 bytes is from https://spl.solana.com/memo#compute-limits
4. Nothing is logged from the initializeProposal in main.ts. So nothing to log there that's not already logged in that function.

PR #101, opened 2024-04-04
Metadata program

Henry-E:
To mention some of the simple design decisions as well. Here's the flow of creating the metadata account and using it
- the DAO treasury automatically signs and creates this account with a CPI call when creating their DAO account in the autocrat program.
- They immediately delegate authority over the metadata account to another account which can sign transactions more easily and update the metadata, since otherwise it would require a DAO for every change to the metadata account. The DAO has authority to reassign the delegate account.
- We track the last updated slot for security reasons or something and to access historical versions easily. Just so people can tell if they got hacked and we can see past revisions by going back to the account data in a given slot. We only update once per slot for the same reason.

Those are the main things we think about and try to account for with this program

----

What's going to be stored in here in practice, basic stuff
- DAO name
- Description
- Socials / Links
All enforcement around keys, values and any schemas will be handled on the UI side.


programs/metadata/src/lib.rs
```diff
@@ -0,0 +1,195 @@
...
+        metadata.last_updated_slot = Clock::get()?.slot;
+        metadata.items = Vec::new();
+        Ok(())
+    }
+
+    pub fn increase_metadata_account_size(
+        _ctx: Context<IncreaseMetadataAccountSize>,
+    ) -> Result<()> {
+        Ok(())
+    }
```
metaproph3t, line 62: Maybe we can do realloc in `ModifyMetadata`, not sure

programs/metadata/src/lib.rs
```diff
@@ -0,0 +1,195 @@
...
+
+#[derive(Accounts)]
+pub struct ModifyMetadata<'info> {
+    #[account(mut, has_one = delegate)]
+    pub metadata: Account<'info, Metadata>,
+    pub delegate: Signer<'info>,
+}
+
+#[error_code]
+pub enum ErrorCode {
```
metaproph3t: I think that this sometimes comes up in tests, which is why the other programs use `AutocratError` and whatnot to differentiate

programs/metadata/src/lib.rs
```diff
@@ -0,0 +1,195 @@
...
+    #[msg("The provided metadata key already exists.")]
+    DuplicateKey,
+    #[msg("Operation cannot be performed on an item updated in the current slot.")]
+    InvalidOperationInCurrentSlot,
```
metaproph3t, line 223: why?

programs/metadata/src/lib.rs
```diff
@@ -0,0 +1,195 @@
...
+            metadata.last_updated_slot = current_slot;
+        } else {
+            msg!("Key: {}", key);
+            return Err(error!(ErrorCode::KeyNotFound));
+        }
+        Ok(())
+    }
+
+    // Instruction to append data to an existing MetadataItem
+    pub fn append_metadata_item(
```
metaproph3t, line 137: what if we took the buffer approach, so instead of writing you write into an offset? so no more append

metaproph3t (commented):
high-level looks great

I wonder if there's a way you can make it so that you don't have any calls into it from autocrat

programs/metadata/src/lib.rs
Henry-E: Will do

programs/metadata/src/lib.rs
Henry-E: This is something nallok and realms dao people mentioned. You don't want there to be multiple updates of a metadata object in the same slot because of security or something about spoofing changes? Plus if there's only one change per item per slot it makes going back and getting the history much cleaner.

programs/metadata/src/lib.rs
Henry-E: Is there a specific issue with append? Using vecs, no zero copy, is much easier than having the UI or the code here figure out where the next empty offset is. Plus we'd have start allocating specific array length to all of the metadata items which are all a bit different.

Henry-E:
- [x] Modify create metadata to load up the dao account and check that it exists before allowing creating a metadata account.

This is instead of adding any code to autocrat. Though we'll need to make sure the UI creates the DAO and the metadata account in the same transaction.

Henry-E:
- [x] Reverting autocrat changes didn't quite work, will need to do another sync with develop and try again later.

Henry-E:
I don't understand why the CI tests keep failing but local is fine. Might need more verbose errors in the tests.

Henry-E:
Devnet details
```
Metadata account exists.
Metadata Account Details:
Metadata Account Public Key: 2nmJLYTKLM9geSYBLAPiFw3UYw1S1SUa9nDTeb1iHNZr
DAO Public Key: 8tanoHEyJEQgaasEkv1DxN6umYNWDotbaEpuzstcEufb
Treasury Public Key: GFBqvgmjZcqSGdHVeHw48UACg6JUBSJMdX94Gf9bhfLG
Delegate Public Key: FutaAyNb3x9HUn1EQNueZJhfy6KCNtAwztvBctoK6JnX

Metadata Items:
1. "description": "An example DAO on Solana"
2. "name": "Example DAO"
3. "twitter": "@exampledao"
4. "discord": "<a fake discord link that triggers the bot somehow>"
```
```
Program Ids
const AUTOCRAT_PROGRAM_ID = new PublicKey(
  "FuTPR6ScKMPHtZFwacq9qrtf9VjscawNEFTb2wSYr1gY"
);
const METADATA_PROGRAM_ID = new PublicKey(
  "DATAwTH3mTx43ekTTYoNXL2Z8EAMTGwENkJab7tKXHok"
);
```
Should be IDLs on chain for both

Still todo
- [ ] figure out why CI tests are failing when local tests are fine
- [ ] Add `init metadata` account instructions to initDao script

PR #104, opened 2024-04-05
Improve constraints

metaproph3t:
Recently, a serious bug was found in the v0.1 version of autocrat that was missed because repeated constraints & verbose naming. Here we slim down the names and abstract away the constraints.

### DIFF
--- a/programs/autocrat_v0/src/lib.rs
+++ b/programs/autocrat_v0/src/lib.rs
@@ -288,12 +288,13 @@ pub mod autocrat_v0 {
     pub fn finalize_proposal(ctx: Context<FinalizeProposal>) -> Result<()> {
         let pass_twap_market = &ctx.accounts.openbook_twap_pass_market;
         let fail_twap_market = &ctx.accounts.openbook_twap_fail_market;
+        let dao = &ctx.accounts.dao;
 
         let proposal = &mut ctx.accounts.proposal;
         let clock = Clock::get()?;
 
         require!(
-            clock.slot >= proposal.slot_enqueued + ctx.accounts.dao.slots_per_proposal,
+            clock.slot >= proposal.slot_enqueued + dao.slots_per_proposal,
             AutocratError::ProposalTooYoung
         );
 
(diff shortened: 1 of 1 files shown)


programs/autocrat_v0/src/lib.rs
```diff
@@ -151,147 +152,84 @@ pub mod autocrat_v0 {
...
-        );
-        require!(
-            ctx.accounts.quote_vault.nonce == (proposal.number as u64 | (1 << 63)),
-            AutocratError::InvalidVaultNonce
-        );
+        for market in [&pass_market, &fail_market] {
+            // The market expires a minimum of 7 days after the end of a 3 day proposal.
+            // Make sure to do final TWAP crank after the proposal period has ended
+            // and before the market expires, or else! Allows for rent retrieval from openbook
+            require!(
```
Henry-E, line 187: It's more of an optional small thing but this might be an opportunity to switch from `require` to `require_gt` or `require_eq`. Just because I think that these other require macros will print out the values on both sides if anything goes wrong, rather than just print require evaluated to false + the error. Very minor thing and not worth risking messing up the existing checks over, so just mentioning it.

programs/autocrat_v0/src/lib.rs
```diff
@@ -151,147 +152,84 @@ pub mod autocrat_v0 {
...
-            openbook_fail_market.quote_mint
-                == ctx.accounts.quote_vault.conditional_on_revert_token_mint,
-            AutocratError::InvalidMarket
-        );
-        require!(
-            openbook_fail_market.time_expiry > current_time + TEN_DAYS_IN_SECONDS,
-            AutocratError::InvalidMarket
-        );
-        require!(
-            openbook_fail_market.seq_num == 0,
```
Henry-E, line 218: This check is missing in the new tests

programs/autocrat_v0/src/lib.rs
```diff
@@ -151,147 +152,84 @@ pub mod autocrat_v0 {
...
+                AutocratError::InvalidMarket
+            );
+
+            require!(
+                market.quote_lot_size == 100, // you can quote in increments of a hundredth of a penny
+                AutocratError::InvalidMarket
+            );
+
+            require!(
+                pass_market.collect_fee_admin == dao.treasury,
```
Henry-E: chatgpt spotted this one. It still has `pass_market` and not `market`

programs/autocrat_v0/src/lib.rs
metaproph3t: intentional, [here's my justification](https://github.com/metaDAOproject/futarchy/pull/104/commits/1cfb3c498084d771d7211492f6096e5b3a33e701)

programs/autocrat_v0/src/lib.rs
metaproph3t: yep, good catch! fixing now

programs/autocrat_v0/src/lib.rs
Henry-E: got it

programs/autocrat_v0/src/lib.rs
metaproph3t: good catch! fixed

PR #107, opened 2024-04-06
Abstract away DAO updates

metaproph3t:
Use a macro to minimize probability of mistypes. OTOH, macros are a more complex rust topic that not everyone knows. So leaving in the drafts for now.

### DIFF
--- a/programs/autocrat_v0/src/lib.rs
+++ b/programs/autocrat_v0/src/lib.rs
@@ -390,39 +390,27 @@ pub mod autocrat_v0 {
     pub fn update_dao(ctx: Context<UpdateDao>, dao_params: UpdateDaoParams) -> Result<()> {
         let dao = &mut ctx.accounts.dao;
 
-        if let Some(pass_threshold_bps) = dao_params.pass_threshold_bps {
-            dao.pass_threshold_bps = pass_threshold_bps;
+        macro_rules! update_fields_if_some {
+            ($dao:expr, $dao_params:expr, $( $field:ident ),+ ) => {
+                $(
+                    if let Some(value) = $dao_params.$field {
+                        $dao.$field = value;
+                    }
+                )+
+            }
         }
 
-        if let Some(base_burn_lamports) = dao_params.base_burn_lamports {
-            dao.base_burn_lamports = base_burn_lamports;
-        }
-
-        if let Some(burn_decay_per_slot_lamports) = dao_params.burn_decay_per_slot_lamports {
-            dao.burn_decay_per_slot_lamports = burn_decay_per_slot_lamports;
-        }
-
-        if let Some(slots_per_proposal) = dao_params.slots_per_proposal {
-            dao.slots_per_proposal = slots_per_proposal;
-        }
-
-        if let Some(market_taker_fee) = dao_params.market_taker_fee {
-            dao.market_taker_fee = market_taker_fee;
-        }
-
-        if let Some(twap_expected_value) = dao_params.twap_expected_value {
-            dao.twap_expected_value = twap_expected_value;
-        }
-
-        if let Some(base_lot_size) = dao_params.base_lot_size {
-            dao.base_lot_size = base_lot_size;
-        }
-
-        if let Some(max_observation_change_per_update_lots) =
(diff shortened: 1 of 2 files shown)


tests/autocratV0.ts
```diff
@@ -1311,39 +1311,34 @@ describe("autocrat_v0", async function () {
...
-          to0,
-          from1: mertdTreasuryUsdcAccount,
-          to1,
-          lamportReceiver: receiver.publicKey,
-        })
-        .instruction();
-
+      const accounts = [
+        {
+          pubkey: mertdDao,
```
R-K-H, line 1316: Any reason this is mertDao?

Henry-E: Think it's just like that to give some flavour to the tests

programs/autocrat_v0/src/lib.rs
```diff
@@ -390,39 +390,27 @@ pub mod autocrat_v0 {
...
-
-        if let Some(burn_decay_per_slot_lamports) = dao_params.burn_decay_per_slot_lamports {
-            dao.burn_decay_per_slot_lamports = burn_decay_per_slot_lamports;
-        }
-
-        if let Some(slots_per_proposal) = dao_params.slots_per_proposal {
-            dao.slots_per_proposal = slots_per_proposal;
-        }
-
-        if let Some(market_taker_fee) = dao_params.market_taker_fee {
```
Henry-E, line 409: Presumably intentionally left out, if so then maybe it should be deleted from the UpdateDaoParams struct. Though not sure if you avoided doing that to stop a breaking change or something

programs/autocrat_v0/src/lib.rs
```diff
@@ -390,39 +390,27 @@ pub mod autocrat_v0 {
...
+                )+
+            }
         }
 
-        if let Some(max_observation_change_per_update_lots) =
-            dao_params.max_observation_change_per_update_lots
-        {
-            dao.max_observation_change_per_update_lots = max_observation_change_per_update_lots;
-        }
+        update_fields_if_some!(
```
Henry-E: The thing I don't like about this macro is that it doesn't separate out the variables on the DAO struct that can get updated from the other vars like dao and dao_parames, it all kind of blends together.

update_field_if_some!(dao, dao_params, pass_threshold_bps);
update_field_if_some!(dao, dao_params, base_burn_lamports);
update_field_if_some!(dao, dao_params, burn_decay_per_slot_lamports);

Above is an alternative but I don't think that level of copypasting is really much of better solution either.

A high effort solution I don't recommend but still mention anyway would be a proc macro that loops over all the var names in UpdateDaoParams and automatically unwraps and tries to assign them to matching var names on the Dao struct. That way you'd never have to update the testing code, it would automatically check all of the params you want to update directly from the UpdateDaoParams struct. Too high effort though and needs a separate crate for the proc macro.

tl;dr LGTM!

Henry-E (changes requested):
Suggest deleting `market_taker_fee` from UpdateDaoParams struct if it was intentionally left out of the `update_dao` function. Unless you want to leave it in there as future proofing so we can update the program without messing up the state. Though I can't imagine a situation we would want to let DAOs set their own market taker fee. 
 https://github.com/metaDAOproject/futarchy/pull/107/files#diff-2b44e9dc4b9dffc5731b087c1ffa4b65bf163b03fb7b917c966182f5b4354a28L530

programs/autocrat_v0/src/lib.rs
metaproph3t: nope, this was accidental. good catch!

programs/autocrat_v0/src/lib.rs
metaproph3t: yeah, agree to both of above. I wish we could do a proc macro, it would be so much cleaner, but probably not worth the marginal complexity.

ended up doing something like you proposed, but without the `dao, dao_params` part: https://github.com/metaDAOproject/futarchy/pull/107/commits/b30aa39d295e351d3a9a5551d1711c1e97cde1e5

Henry-E (commented):
Looks good

PR #108, opened 2024-04-07
Feat/improve vault security

metaproph3t:
Do the same thing that #104 did, but for the conditional vault logic.

### DIFF
--- a/programs/conditional_vault/src/lib.rs
+++ b/programs/conditional_vault/src/lib.rs
@@ -75,16 +75,19 @@ pub mod conditional_vault {
     ) -> Result<()> {
         let vault = &mut ctx.accounts.vault;
 
-        vault.status = VaultStatus::Active;
-        vault.settlement_authority = settlement_authority;
-        vault.underlying_token_mint = ctx.accounts.underlying_token_mint.key();
-        vault.nonce = nonce;
-        vault.underlying_token_account = ctx.accounts.vault_underlying_token_account.key();
-        vault.conditional_on_finalize_token_mint =
-            ctx.accounts.conditional_on_finalize_token_mint.key();
-        vault.conditional_on_revert_token_mint =
-            ctx.accounts.conditional_on_revert_token_mint.key();
-        vault.pda_bump = *ctx.bumps.get("vault").unwrap();
+        vault.set_inner(ConditionalVault {
+            status: VaultStatus::Active,
+            settlement_authority,
+            underlying_token_mint: ctx.accounts.underlying_token_mint.key(),
+            nonce,
+            underlying_token_account: ctx.accounts.vault_underlying_token_account.key(),
+            conditional_on_finalize_token_mint: ctx
+                .accounts
+                .conditional_on_finalize_token_mint
+                .key(),
+            conditional_on_revert_token_mint: ctx.accounts.conditional_on_revert_token_mint.key(),
+            pda_bump: *ctx.bumps.get("vault").unwrap(),
+        });
 
         Ok(())
     }
--- a/programs/conditional_vault/src/lib.rs
+++ b/programs/conditional_vault/src/lib.rs
@@ -278,11 +278,11 @@ pub mod conditional_vault {
     pub fn mint_conditional_tokens(ctx: Context<MintConditionalTokens>, amount: u64) -> Result<()> {
         let accs = &ctx.accounts;
 
+        let pre_vault_underlying_balance = accs.vault_underlying_token_account.amount;
         let pre_user_conditional_on_finalize_balance =
             accs.user_conditional_on_finalize_token_account.amount;
         let pre_user_conditional_on_revert_balance =
             accs.user_conditional_on_revert_token_account.amount;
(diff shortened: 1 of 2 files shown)


programs/conditional_vault/src/lib.rs
```diff
@@ -75,16 +75,19 @@ pub mod conditional_vault {
...
-        vault.settlement_authority = settlement_authority;
-        vault.underlying_token_mint = ctx.accounts.underlying_token_mint.key();
-        vault.nonce = nonce;
-        vault.underlying_token_account = ctx.accounts.vault_underlying_token_account.key();
-        vault.conditional_on_finalize_token_mint =
-            ctx.accounts.conditional_on_finalize_token_mint.key();
-        vault.conditional_on_revert_token_mint =
-            ctx.accounts.conditional_on_revert_token_mint.key();
-        vault.pda_bump = *ctx.bumps.get("vault").unwrap();
+        vault.set_inner(ConditionalVault {
```
Henry-E, line 78: Looks good

programs/conditional_vault/src/lib.rs
```diff
@@ -106,41 +109,48 @@ pub mod conditional_vault {
         let on_finalize_token_symbol = format!("p{}", underlying_token_symbol);
```
Henry-E, line 109: @jshiohaha Random question about token metadata. If the token symbol is just `p{}` so `pMeta` won't that make it hard to tell apart / search for the different proposal tokens if there are 10s of `pMeta`s, even if the toplevel metadata name is different? Aside from aesthetics is there a downside to having the ticker as pMeta_{proposal_number}?

programs/conditional_vault/src/lib.rs
```diff
@@ -106,41 +109,48 @@ pub mod conditional_vault {
...
-                seller_fee_basis_points: 0,
-                creators: None,
-                collection: None,
-                uses: None,
-            },
-            false,
-            true,
-            None,
-        )?;
+        for (symbol, uri, metadata, mint) in [
```
Henry-E, line 112: less code, looks good

programs/conditional_vault/src/lib.rs
```diff
@@ -369,73 +378,65 @@ pub mod conditional_vault {
...
-                    accs.token_program.to_account_info(),
-                    Transfer {
-                        from: accs.vault_underlying_token_account.to_account_info(),
-                        to: accs.user_underlying_token_account.to_account_info(),
-                        authority: accs.vault.to_account_info(),
-                    },
-                    signer,
-                ),
-                conditional_on_revert_balance,
+                user_conditional_token_account.amount,
```
Henry-E, line 417: It's probably not necessary but it might have been possible to reuse the pre_conditional_balances here.

programs/conditional_vault/src/lib.rs
```diff
@@ -265,17 +275,22 @@ pub mod conditional_vault {
         Ok(())
     }
 
-    pub fn mint_conditional_tokens(ctx: Context<MintConditionalTokens>, amount: u64) -> Result<()> {
+    pub fn mint_conditional_tokens(ctx: Context<InteractWithVault>, amount: u64) -> Result<()> {
```
Henry-E, line 278: It probably doesn't matter and it wasn't in the original account struct but I wonder if there's any issues or downside with limiting minting to when the market is active?

Henry-E:
LGTM

programs/conditional_vault/src/lib.rs
jshiohaha: That sounds like what I'd expect given the current design. But, I actually saw this name + symbol implementation as a benefit, essentially like a primary and secondary sort/filter parameter — e.g. you could filter on all `pMETA` first and then find subset you are interested in based on the proposal count. Or, you could start at the name level. Although, I could be misunderstanding the eventual query patterns we will use.

If there's a reason to make both unique, I'm certainly not against it! Although, ticker like `pMETA_{proposal_number}` doesn't look very nice IMO 😅

Henry-E: Sure, the underscore does mess with the aesthetics. Probably best to leave as is for now and just see how annoying / not annoying it is to find tokens once there's been a few proposals up using the new contracts.

programs/conditional_vault/src/lib.rs
metaproph3t: I think you want people to be able to capture arbs in expired markets when they exist

programs/conditional_vault/src/lib.rs
metaproph3t: thought about it, but I'd next an extra param to the for loop

programs/conditional_vault/src/lib.rs
Henry-E: Fair enough

programs/conditional_vault/src/lib.rs
Henry-E: It would yeh, but no worries

Issue #112, opened 2024-04-09
Start `proposal_count` at 0, first proposal should be 1


metaproph3t:
Normies don't think in 0-indexed terms, so the first proposal should be 'proposal 1.' But we should also start the proposal count at 0. Should be a smol change.

metaproph3t:
Addressed with #113

PR #113, opened 2024-04-09
Start `proposal_count` at 0 and first proposal to 1

metaproph3t:
Address #112


metaproph3t:
FYI @R-K-H

Henry-E:
The intrusive thoughts won

PR #144, opened 2024-04-18
feat: legacy support, initialize guardrails, finalize

R-K-H:
I've updated this for review in DEVELOP.

This includes
- new tokens for the management of DAOs on Devnet
- cleaned up initialize with guard rails and checks
- finalize function which is fairly robust for most finalizations
- output of txn for initialize dao
- dao's pub key output for storage

### DIFF
--- a/scripts/initializeDao.ts
+++ b/scripts/initializeDao.ts
@@ -1,14 +1,57 @@
-import { autocratProgram, initializeDAO } from "./main";
+import { provider, DAO, initializeDAO, BASE_TOKEN, QUOTE_TOKEN } from "./main";
 import * as anchor from "@coral-xyz/anchor";
+// @ts-ignore
+import * as token from "@solana/spl-token";
 
-const { PublicKey, Keypair, SystemProgram } = anchor.web3;
-const { BN, Program } = anchor;
+const { Keypair } = anchor.web3;
 
-const META = new PublicKey("METADDFL6wWMWEoKTFJwcThTbUmtarRJZjRpzUvkxhr");
-const USDC = new PublicKey("EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v");
+const TOKEN = BASE_TOKEN;
+const USDC = QUOTE_TOKEN;
 
 async function main() {
-  await initializeDAO(META, USDC);
+  console.log('Initializing Futarchic DAO')
+  console.log(`Base Token ${BASE_TOKEN.toString()}`)
+  console.log(`Quote Token ${QUOTE_TOKEN.toString()}`)
+  if(DAO.publicKey) {
+    console.log("ALREADY INITIALIZED DAO WITH THIS NAME!!")
+    return;
+  }
+  // Generate KP for use with DAO
+  const daoKP = Keypair.generate();
+
+  // Fetch data about our token...
+  const tokenData = await token.getMint(provider.connection, TOKEN);
+  const decimals = tokenData.decimals;
+  console.log(`Fetched Token Decimals: ${decimals}`)
+
+  // TODO: Query JUP for swap amount to get value
+  const tokenPrice = 0.0008 // Rounded to nearest
+
+  // Get a price between 0 and 10 whole value
+  // Example: 1 META @ $400 would be 0.01 as that's $4
+  // Example: 1 DEAN @ $0.000795 would be 10_000 as that's $7.81
+  const lotOfToken = 10_000 // This is token quantity to get between $0-$10
+  const twapExpectedDollar = tokenPrice * lotOfToken
+  if (twapExpectedDollar > 10) {
+    console.log(`Check your numbers, as $${twapExpectedDollar} is larger than $10`)
+    return;
+  }
+  
+  const baseLotSize = lotOfToken * (10 ** decimals);
+  // Take the whole dollar value * 10_000
+  const twapExpectedValue = (twapExpectedDollar) * 10_000; // 10_000 is hundredths of pennies
+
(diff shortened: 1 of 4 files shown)


scripts/initializeDao.ts
```diff
@@ -1,14 +1,57 @@
...
+  }
+  // Generate KP for use with DAO
+  const daoKP = Keypair.generate();
+
+  // Fetch data about our token...
+  const tokenData = await token.getMint(provider.connection, TOKEN);
+  const decimals = tokenData.decimals;
+  console.log(`Fetched Token Decimals: ${decimals}`)
+
+  // TODO: Query JUP for swap amount to get value
```
R-K-H, line 27: Would be nice to do something JUP and actually feed users this..

scripts/initializeDao.ts
```diff
@@ -1,14 +1,57 @@
...
+  const tokenPrice = 0.0008 // Rounded to nearest
+
+  // Get a price between 0 and 10 whole value
+  // Example: 1 META @ $400 would be 0.01 as that's $4
+  // Example: 1 DEAN @ $0.000795 would be 10_000 as that's $7.81
+  const lotOfToken = 10_000 // This is token quantity to get between $0-$10
```
R-K-H, line 33: This should be calculated, but I've exhausted my capacity to think / understand what would be best.

scripts/main.ts
```diff
@@ -37,34 +37,200 @@ const AutocratIDL: AutocratV0 = require("../target/idl/autocrat_v0.json");
...
+// These programs don't always have upgrades, therefore you want to
+// use the matching version or one version less.
+// eg. v0.3 autocrat uses v0.2 twap version, or v0.1 uses v0 conditional vault
+// NOTE: The only exception is the migrator, which extends one beyond
+// eg. v0 autocrat uses v0.1 autocrat migrator
+// TODO: We need some logic to loop through and find highest version number and then step back one
+const CONDITIONAL_VAUT = CONDITIONAL_VAULT_VERSIONS.find(
+  (program) => 
+    // NOTE: This is a quick and dirty way
+    program.version === useVersion || program.version === useVersion - 1 || program.version
```
R-K-H, line 204: This has a bug, in that it is looping from index 0, and really we want the greatest if we can find it, not the first one it finds.

scripts/main.ts
```diff
@@ -37,34 +37,200 @@ const AutocratIDL: AutocratV0 = require("../target/idl/autocrat_v0.json");
...
+)
+const OPENBOOK_TWAP = OPENBOOK_TWAP_VERSIONS.find(
+  (program) => 
+    // NOTE: This is a quick and dirty way
+    program.version === useVersion || program.version === useVersion - 1 || program.version
```
R-K-H, line 209: This has a bug, in that it is looping from index 0, and really we want the greatest if we can find it, not the first one it finds.

scripts/main.ts
```diff
@@ -37,34 +37,200 @@ const AutocratIDL: AutocratV0 = require("../target/idl/autocrat_v0.json");
...
+)
+const AUTOCRAT_MIGRATOR = AUTOCRAT_MIGRATOR_VERSIONS.find(
+  // TODO: Hack to get this to play nice...
+  (program) => program.version === useVersion + 1 || program.version === useVersion - 1
```
R-K-H, line 213: This has a bug, in that it is looping from index 0, and really we want the greatest if we can find it, not the first one it finds.

scripts/main.ts
```diff
@@ -298,104 +482,97 @@ async function initializeVault(
...
+    microLamports: 1000
+  });
+
+  const accounts = storedProposal.instruction.accounts
+  const program = storedProposal.instruction.programId
+
+  // Set this up for certain actions with the treasury such that the 
+  // second signer isn't really a signer..
+  try {
+    accounts[2].isSigner = false
```
R-K-H, line 543: I need this sometimes, and other times not, I have no idea why.

scripts/uploadOffchainMetadata.ts
```diff
@@ -36,14 +36,14 @@ const payer = provider.wallet["payer"];
 const umi = createUmi(provider.connection);
 umi.use(keypairIdentity(payer));
 
-type AcceptedVaultTokens = "META" | "USDC";
-type ConditionalToken = "pMETA" | "fMETA" | "pUSDC" | "fUSDC";
+type AcceptedVaultTokens = "META" | "USDC" | "FUTURE" | "TOKE" | "DEAN";
```
R-K-H, line 39: Ideally we have a better or shared lib or SOMETHING else other than adding these.

scripts/main.ts
```diff
@@ -37,34 +37,200 @@ const AutocratIDL: AutocratV0 = require("../target/idl/autocrat_v0.json");
...
 );
-export const DEVNET_USDC = new PublicKey(
+// Base Tokens
+// Meta USDC (created for use in our contracts)
+export const DEVNET_MUSDC = new PublicKey(
   "B9CZDrwg7d34MiPiWoUSmddriCtQB5eB2h9EUSDHt48b"
+)
+// Circle Devnet USDC (from faucet)
+export const DEVNET_USDC = new PublicKey(
+  "4zMMC9srt5Ri5X14GAgXhaHii3GnPAEERYPJgZJDncDU"
```
R-K-H, line 66: @jshiohaha this token is just normal token program but USDC which comes from the faucet for USDC https://faucet.circle.com/ is there a way to use original tokenkeg or is that data pulled from the OLD token list etc? It shouldn't be priority, however just was curious what happens if we don't use metaplex?

PR #235, opened 2024-06-25
feat: create proposal tx method

LukasDeco:


metaproph3t:
what's the purpose of this change?

LukasDeco:
> what's the purpose of this change?

Actually I need to make a change here. But the purpose is to just create a method for proposal initialization. I will need a proposal that will return an array of the transactions in the correct order and that we can sign them and send them to the jito bundle. That's what I'm working on here essentially but... this doesn't do that obviously because it uses .rpc(). This I guess will just work on devnet and won't really work on mainnet too well so... I might just mark this as draft for now.

LukasDeco:
@metaproph3t this is ready now but not sure why anchor test is failing?

metaproph3t (approved):
LGTM

PR #237, opened 2024-07-01
Check in production that Proph3t is the one adding metadata

metaproph3t:


Henry-E:
More security related PRs?

R-K-H:
The creation of the metadata is malformed. You cannot put the image URL in the URI field and be able to process it with the standard.

Issue #95, opened 2024-04-03
Move from `justfile` to shell scripts


metaproph3t:
Just is annoying because most computers don't have it by default and it's non-trivial to install via apt. We should move to a shell script to make it easier to run tests.

metaproph3t:
Resolved via #243

PR #243, opened 2024-08-16
removed `justfile` - resolving issue 95

ith-harvey:
Resolved issue https://github.com/metaDAOproject/futarchy/issues/95
- removed `justfile`
- moved all test CI commands to `test.sh`
- decided with shell vs bash due to higher number of comps having shell (don't love how function names have `_` instead of `-`)
- cleaned up `RUN_INSTRUCTIONS.md`
- tested all methods successfully aside from `deploy()`, `upgrade()`, `upgrade_idl()` as `solana` CI was saying some of the args were incompatible. Decided to leave them as is since it could have been a config issue on my end.


metaproph3t:
Nice - thank you sir!

PR #244, opened 2024-08-23
Arbitrary conditional vault

metaproph3t:


Henry-E:
I will try to give this a read through when i get a chance if that's helpful

programs/conditional_vault/src/state/question.rs
```diff
@@ -0,0 +1,39 @@
...
+/// also have questions with scalar outcomes. For example, the question "how
+/// effective will the grant committee deem this grant" could have two outcomes:
+/// "ineffective" and "effective." If the grant committee deems the grant 70%
+/// effective, the "effective" outcome would resolve to 0.7 and the "ineffective"
+/// outcome would resolve to 0.3.
+///
+/// Once resolved, the sum of all outcome resolutions is exactly 1.
+#[account]
+pub struct Question {
+    pub question_id: [u8; 32],
```
0xbigz, line 25: 32 characters might be kinda smol, if its just an ID (and not characters) why do it like this?

metaproph3t: it's supposed to be a hash

programs/conditional_vault/src/instructions/redeem_tokens.rs
```diff
@@ -0,0 +1,127 @@
...
+            // this is safe because we check that every conditional mint is a part of the vault
+            let payout_index = vault
+                .conditional_token_mints
+                .iter()
+                .position(|mint| mint == &conditional_mint.key())
+                .unwrap();
+
+            total_redeemable += ((user_conditional_token_account.amount as u128
+                * question.payout_numerators[payout_index] as u128)
+                / question.payout_denominator as u128) as u64;
```
0xbigz, line 58: doing a mix of Solana `Context` and (unsafe) math logic within one function. would be great to split up math and state modification logic from solana account loader stuff. this makes rust tests much easier to write and lets you fuzz more easily

PR #246, opened 2024-08-29
v0.4

metaproph3t:
see this new PR @0xbigz @Henry-E


Henry-E:
Reviewing it now. It would be useful to get a short high level overview of some of the changes and the new structure.

programs/autocrat/src/instructions/initialize_proposal.rs
```diff
@@ -25,28 +25,34 @@ pub struct InitializeProposal<'info> {
...
         constraint = base_vault.underlying_token_mint == dao.token_mint,
-        constraint = base_vault.settlement_authority == proposal.key() @ AutocratError::InvalidSettlementAuthority,
+        has_one = question,
+        // constraint = base_vault.settlement_authority == proposal.key() @ AutocratError::InvalidSettlementAuthority,
     )]
     pub base_vault: Account<'info, ConditionalVaultAccount>,
     #[account(
-        constraint = pass_amm.base_mint == base_vault.conditional_on_finalize_token_mint,
-        constraint = pass_amm.quote_mint == quote_vault.conditional_on_finalize_token_mint,
+        constraint = pass_amm.base_mint == base_vault.conditional_token_mints[1],
```
Henry-E: Instead of using 0 and 1 everywhere i'm wondering if it would be better / possible to use an enum. So it's clear this is referring to pass / fail.

programs/conditional_vault/src/error.rs
```diff
@@ -2,14 +2,34 @@ use super::*;
...
     #[msg("Insufficient underlying token balance to mint this amount of conditional tokens")]
     InsufficientUnderlyingTokens,
+    #[msg("Insufficient conditional token balance to merge this `amount`")]
+    InsufficientConditionalTokens,
     #[msg("This `vault_underlying_token_account` is not this vault's `underlying_token_account`")]
     InvalidVaultUnderlyingTokenAccount,
     #[msg("This conditional token mint is not this vault's conditional token mint")]
     InvalidConditionalTokenMint,
-    #[msg("Vault needs to be settled as finalized before users can redeem conditional tokens for underlying tokens")]
+    #[msg("Question needs to be resolved before users can redeem conditional tokens for underlying tokens")]
```
Henry-E, line 15: It's nitpicking but the idea passing or failing a question is funny / odd. 

This might be the wrong place for this comment. In autocrat it's more odd since we still retain the pass / fail model (thankfully)

programs/conditional_vault/src/instructions/add_metadata_to_conditional_tokens.rs
```diff
@@ -8,53 +8,41 @@ pub mod proph3t_deployer {
...
-    pub underlying_token_mint: Account<'info, Mint>,
-    pub underlying_token_metadata: Account<'info, MetadataAccount>,
-    #[account(
-        mut,
-        mint::authority = vault,
-        mint::freeze_authority = vault,
-        mint::decimals = underlying_token_mint.decimals
-    )]
-    pub conditional_on_finalize_token_mint: Account<'info, Mint>,
+    pub vault: Account<'info, ConditionalVault>,
```
Henry-E, line 22: How come the vault no longer has to be connected to the conditional token mint via a `has_one` or other check?

Henry-E: the `mint::authority = vault,` check on `conditional_token_mint` is probably enough maybe? It's still not like a fully exclusive guarantee but whatever.

programs/conditional_vault/src/instructions/redeem_tokens.rs
```diff
@@ -0,0 +1,127 @@
...
+
+        let (mut conditional_token_mints, mut user_conditional_token_accounts) =
+            Self::get_mints_and_user_token_accounts(&ctx)?;
+
+        // calculate the expected future supplies of the conditional token mints
+        // as current supply - user balance
+        let expected_future_supplies: Vec<u64> = conditional_token_mints
+            .iter()
+            .zip(user_conditional_token_accounts.iter())
+            .map(|(mint, account)| mint.supply - account.amount)
```
Henry-E, line 24: does this need to be checked, i assume if someone is trying to withdraw `account.amount` > `mint.supply` it will error somewhere along the line

programs/conditional_vault/src/instructions/resolve_question.rs
```diff
@@ -0,0 +1,33 @@
...
+    pub fn handle(ctx: Context<Self>, args: ResolveQuestionArgs) -> Result<()> {
+        let question = &mut ctx.accounts.question;
+
+        require_eq!(
+            args.payout_numerators.len(),
+            question.num_outcomes(),
+            VaultError::InvalidNumPayoutNumerators
+        );
+
+        question.payout_denominator = args.payout_numerators.iter().sum();
```
Henry-E, line 29: what's going on here?

programs/autocrat/src/instructions/finalize_proposal.rs
```diff
@@ -147,38 +152,41 @@ impl FinalizeProposal<'_> {
             .saturating_mul(MAX_BPS.saturating_add(dao.pass_threshold_bps).into())
             / MAX_BPS as u128;
 
-        let (new_proposal_state, new_vault_state) = if pass_market_twap > threshold {
-            (ProposalState::Passed, VaultStatus::Finalized)
+        let (new_proposal_state, payout_numerators) = if pass_market_twap > threshold {
+            (ProposalState::Passed, vec![0, 1])
```
Henry-E, line 152: ok, i see now that these are the "numerators" that say the proposal 100% passed or 100% failed. On the surface of it, not super intuitive. Don't have an immediate suggestion though.

programs/conditional_vault/src/state/question.rs
```diff
@@ -0,0 +1,39 @@
...
+///
+/// Questions have 2 or more possible outcomes. For a question like "will this
+/// proposal pass," the outcomes are "yes" and "no." For a question like "who
+/// will be hired," the outcomes could be "Alice," "Bob," and "neither."
+///
+/// Outcomes resolve to a number between 0 and 1. Binary questions like "will
+/// this proposal pass" have outcomes that resolve to exactly 0 or 1. You can
+/// also have questions with scalar outcomes. For example, the question "how
+/// effective will the grant committee deem this grant" could have two outcomes:
+/// "ineffective" and "effective." If the grant committee deems the grant 70%
```
Henry-E, line 18: And that means the people who hold `effective` tokens get 70% of the supply in the conditional token vaults?

programs/autocrat/src/instructions/initialize_proposal.rs
metaproph3t: 100%, this is a great idea

programs/conditional_vault/src/instructions/resolve_question.rs
0xbigz: `payout_numerators` vector is relative share of tokens to be distributed between different answers

sum of all the numerators is the denominator (suppose to guarantee sum to 100%)

theres likely going to be dust / rounding errors from this (and should be handled)

Henry-E: ok wow, i need to review your big write up on the structure. That seems hard to bet on. If something can be 70% considered highly effective and 30% considered moderately effective. Like maybe in most cases it should just resolve to a single outcome. But having to balance the idea it might be a mix of outcomes, it's a foreign idea at the moment.

programs/conditional_vault/src/instructions/resolve_question.rs
```diff
@@ -0,0 +1,33 @@
...
+impl ResolveQuestion<'_> {
+    pub fn handle(ctx: Context<Self>, args: ResolveQuestionArgs) -> Result<()> {
+        let question = &mut ctx.accounts.question;
+
+        require_eq!(
+            args.payout_numerators.len(),
+            question.num_outcomes(),
+            VaultError::InvalidNumPayoutNumerators
+        );
+
```
92GC, line 25: I think we require a check that there is a winning outcome. 

// Find the maximum value
let max_value = *args.payout_numerators.iter().max().unwrap_or(&0);

// Check if the maximum value is unique
if args.payout_numerators.iter().filter(|&&x| x == max_value).count() != 1 {
    return Err(VaultError::NonUniquePayout);
}

programs/conditional_vault/src/state/question.rs
```diff
@@ -0,0 +1,39 @@
...
+/// effective, the "effective" outcome would resolve to 0.7 and the "ineffective"
```
92GC, line 19: The question is made of u32, where are these transformed into integers

    pub payout_numerators: Vec<u32>,
    pub payout_denominator: u32,

programs/conditional_vault/src/instructions/resolve_question.rs
92GC, line 28: Should we check that the options add up to 100%:

        require_eq!(
            question.payout_denominator,
            args.payout_numerators.iter().sum(),
            VaultError::InvalidNumPayoutNumerators
        );

programs/conditional_vault/src/state/question.rs
```diff
@@ -0,0 +1,39 @@
...
+/// Questions represent statements about future events.
+///
+/// These statements include:
+/// - "Will this proposal pass?"
+/// - "Who, if anyone, will be hired?"
+/// - "How effective will the grant committee deem this grant?"
+///
+/// Questions have 2 or more possible outcomes. For a question like "will this
+/// proposal pass," the outcomes are "yes" and "no." For a question like "who
+/// will be hired," the outcomes could be "Alice," "Bob," and "neither."
```
92GC, line 12: 1) Alice and Bob could be hired, so this is a false choice fallacy. Proposers need to be cautious with mutually exclusive outcomes. If they are in the same proposal we must have markets for the Cartesian product of possible states.

2) Maybe every multi-option decision market should have an "other" option as a catch-all. Or state clearly state in the proposal that if another outcome happens all markets will be treated as non-canonical.

metaproph3t:
> Reviewing it now. It would be useful to get a short high level overview of some of the changes and the new structure.

High-level: abstractify the conditional vault to allow for multiple options and variable payouts for each option. We do this via having "questions" which have "outcomes," which themselves have "payouts." It's pretty similar to the way that [Gnosis' conditional token framework](https://github.com/gnosis/conditional-tokens-contracts) works.

programs/autocrat/src/instructions/initialize_proposal.rs
metaproph3t: Addressed by 1b68f4b68dfa06fe0a670918ee7199471733672b

programs/conditional_vault/src/error.rs
metaproph3t: In this case, the question is "Will proposal x pass or fail?" and you're resolving it to "Pass" or "Fail"

programs/conditional_vault/src/instructions/add_metadata_to_conditional_tokens.rs
metaproph3t: these checks are pretty pointless until we're able to verify URIs on-chain so that we can make this permissionless

programs/conditional_vault/src/instructions/redeem_tokens.rs
metaproph3t: I was gonna type "it doesn't hurt to check," but it does because you're adding complexity

I agree this should never be triggered, it's more of a sanity check on myself. my bayesian network tells me it's worth keeping in there :shrug:

programs/conditional_vault/src/instructions/resolve_question.rs
metaproph3t: this allows you to do it on any metric tho, that's the nice thing. like if some wants to lose weight but is not sure which diet plan is best, they can create conditional markets on what would be their weight be dependent on which plan they eat

metaproph3t: > `payout_numerators` vector is relative share of tokens to be distributed between different answers
> 
> sum of all the numerators is the denominator (suppose to guarantee sum to 100%)
> 
> theres likely going to be dust / rounding errors from this (and should be handled)

we're always rounding in the vault's favor, so there will be dust left in the vault and that's okay

programs/conditional_vault/src/state/question.rs
metaproph3t: yes

programs/conditional_vault/src/instructions/resolve_question.rs
metaproph3t: this is implicit because we're setting question.payout_denominator equal to this sum, so check not needed

programs/conditional_vault/src/state/question.rs
metaproph3t: 1. I'm saying it *could* be the outcomes (in the case where you only have enough money to hire one, say), not that it always would
2. what would be an example of where you'd want an 'other'?

programs/conditional_vault/src/instructions/redeem_tokens.rs
Henry-E: For sure, just checking in case there was some other reason for it perhaps. Sounds good

programs/conditional_vault/src/instructions/redeem_tokens.rs
```diff
@@ -0,0 +1,127 @@
...
+        let signer = &[&seeds[..]];
+
+        let user_underlying_balance_before = accs.user_underlying_token_account.amount;
+        let vault_underlying_balance_before = accs.vault_underlying_token_account.amount;
+        // safe because there is always at least two conditional tokens and thus
+        // at least two user conditional token accounts
+        let max_redeemable = user_conditional_token_accounts
+            .iter()
+            .map(|account| account.amount)
+            .max()
```
Henry-E, line 40: Pretty sure the maximum redeemable amount would be the minimum of all the account values?

Henry-E: Never mind, I thought this was the merge instruction

Henry-E:
So what's the intention for the multi-outcome markets here?
- The conditional vaults can be used to create conditional tokens
- There's an oracle that determines the outcome
- The AMM still only deals with TWAP proposals
Where does trading for multi-outcome markets actually happen? Is the UI just going to plug into a bunch of regular meteora AMMs or is there another program still to come?

metaproph3t:
> So what's the intention for the multi-outcome markets here?
> 
>     * The conditional vaults can be used to create conditional tokens
> 
>     * There's an oracle that determines the outcome
> 
>     * The AMM still only deals with TWAP proposals
>       Where does trading for multi-outcome markets actually happen? Is the UI just going to plug into a bunch of regular meteora AMMs or is there another program still to come?

Each outcome will have its own AMM markets, so we don't need to change the amm program to support this. We will need to change autocrat to handle multiple markets (it'll need to receive all of them and pick the option with the highest TWAP).

programs/conditional_vault/src/state/question.rs
92GC: I was just basing it off etiquette in Reddit polls. But here's a thought experiment:

A DAO has a decision market on whether a grant will be considered effective or not, by a committee, in six months. If the committee does not respond within seven months maybe the 'other' option should be declared the fallback outcome by the oracle and (some of?) the proposers’ AMM liquidity should be kept by the DAO as a penalty.

On further thought asking the question correctly solves this. E.g. Will the committee decide this grant is effective in 6 months. All other outcomes are encapsulated by the no. See [Drift's "Will Donald Trump win the 2024 Presidential Election?" [Yes] or [No]](https://app.drift.trade/bet/TRUMP-WIN-2024-BET)

However, if there are >2 distinct outcomes that a proposer wants to be traded, I think "other" has to come into play. See Polymarket election winner:

image

This applies to DAOs more so since, with decentralised organizations and teams, there is a risk of message delivery failure from the committees to the market oracle.

Not sure if nerd sniping or helping here, feel free to resolve the thread.

92GC:
> Each outcome will have its own AMM markets, so we don't need to change the amm program to support this. 

If we ignored constraints on dev time, IMO it would be better if we used [FPMMs](https://docs.polymarket.com/#fpmms) like Polymarket and Gnosis. Using separate AMMs is leaky. If I buy Pass I am explicitly selling Fail here. Arb bots will be able to scalp across AMMs here. I guess pro traders can use Jito bundles though. 

Using an FPMM would also get rid of the leak in the proposers' liquidity from arbs with the spot market. Though, these arbs are good for the DAO metrics and revenue.

This is audited code for a multi-asset AMM: [https://github.com/igneous-labs/s](https://github.com/igneous-labs/s)

metaproph3t:
> > Each outcome will have its own AMM markets, so we don't need to change the amm program to support this.
> 
> If we ignored constraints on dev time, IMO it would be better if we used [FPMMs](https://docs.polymarket.com/#fpmms) like Polymarket and Gnosis. Using separate AMMs is leaky. If I buy Pass I am explicitly selling Fail here. Arb bots will be able to scalp across AMMs here. I guess pro traders can use Jito bundles though.
> 
> Using an FPMM would also get rid of the leak in the proposers' liquidity from arbs with the spot market. Though, these arbs are good for the DAO metrics and revenue.

The AMM is actually already a FPMM (more commonly referred to as a CPMM). How would using one stop value leakage through spot market arbing?

92GC:
> How would using one stop value leakage through spot market arbing?

Sorry, this doesn't fix arbs with spot. There won't be any arbs with spot for betting markets with PassUSDC/FailUSDC tokens.

Though for DAO related decisions a PassDaoToken / FailDAOToken betting market is more capital efficient for holders.

programs/autocrat/src/instructions/finalize_proposal.rs
```diff
@@ -1,10 +1,12 @@
...
+
 use super::*;
 
 #[derive(Accounts)]
 pub struct FinalizeProposal<'info> {
     #[account(mut,
-        has_one = base_vault,
-        has_one = quote_vault,
+        // has_one = base_vault,
+        // has_one = quote_vault,
```
robre: shouldn't there be a `has_one = question` now?

programs/conditional_vault/src/instructions/initialize_conditional_vault.rs
```diff
@@ -53,26 +31,67 @@ pub struct InitializeConditionalVault<'info> {
...
+            let seeds = &[
+                b"conditional_token",
+                vault_key.as_ref(),
+                &[i as u8],
+                &[pda_bump],
+            ];
+            let signer = &[&seeds[..]];
+            let cpi_ctx = CpiContext::new_with_signer(cpi_program, cpi_accounts, signer);
+
+            system_program::create_account(
```
robre: https://github.com/solana-labs/solana/blob/27eff8408b7223bb3c4ab70523f8a8dca3ca6645/programs/system/src/system_processor.rs#L161
when accounts already have lamports, create_account will fail
attackers may send a lamport to your conditional accounts and make initialization impossible.

programs/autocrat/src/instructions/finalize_proposal.rs
robre: this might be a critical issue.
1. create proposal with question A, and let it pass
2. create question B with the same proposal as oracle
3. call finalize_proposal with the proposal, but with question B
4. question B will now be resolved, and the proposal will be blocked from resolving with question A because the state is now passed/failed

programs/conditional_vault/src/instructions/merge_tokens.rs
```diff
@@ -0,0 +1,106 @@
...
+                    authority: accs.vault.to_account_info(),
+                },
+                signer,
+            ),
+            amount,
+        )?;
+
+        ctx.accounts.user_underlying_token_account.reload()?;
+        ctx.accounts.vault_underlying_token_account.reload()?;
+
```
robre, line 67: nice, I like the meticulous checks here

programs/conditional_vault/src/instructions/redeem_tokens.rs
```diff
@@ -0,0 +1,127 @@
...
+            .zip(user_conditional_token_accounts.iter())
+        {
+            // this is safe because we check that every conditional mint is a part of the vault
+            let payout_index = vault
+                .conditional_token_mints
+                .iter()
+                .position(|mint| mint == &conditional_mint.key())
+                .unwrap();
+
+            total_redeemable += ((user_conditional_token_account.amount as u128
```
robre, line 56: this logic means that we're rounding down (e.g. 1000 * 100/900 -> 111). This means the vault will end up with some dust

programs/conditional_vault/src/instructions/redeem_tokens.rs
```diff
@@ -0,0 +1,127 @@
...
+
+        require_eq!(
+            ctx.accounts.vault_underlying_token_account.amount,
+                vault_underlying_balance_before - total_redeemable,
+                VaultError::AssertFailed
+        );
+
+        for acc in user_conditional_token_accounts.iter_mut() {
+            acc.reload()?;
+            require_eq!(acc.amount, 0, VaultError::AssertFailed);
```
robre, line 105: might as well close the accounts

programs/conditional_vault/src/state/conditional_vault.rs
```diff
@@ -9,27 +9,60 @@ pub enum VaultStatus {
...
+
+        let max_possible_liability = if !question.is_resolved() {
+            // safe because conditional_token_supplies is non-empty
+            *conditional_token_supplies.iter().max().unwrap()
+        } else {
+            conditional_token_supplies
+                .iter()
+                .enumerate()
+                .map(|(i, supply)| {
+                    *supply * question.payout_numerators[i] as u64
```
robre: in redeem_tokens.rs, you cast to u128 for safe multiplication. I'd recommend doing it here as well

robre (changes requested):
I've reviewed the programs, and added some comments. Most importantly you should add a check when resolving amms/questions (has_one = question), as it may be a critical bug.
Otherwise looks mostly quite solid, just added tiny comments

robre:
ah please note my comments often refer to the code before/after what's shown by github as the context...

programs/conditional_vault/src/instructions/redeem_tokens.rs
```diff
@@ -0,0 +1,141 @@
...
+            // this is safe because we check that every conditional mint is a part of the vault
+            let payout_index = vault
+                .conditional_token_mints
+                .iter()
+                .position(|mint| mint == &conditional_mint.key())
+                .unwrap();
+
+            total_redeemable += ((user_conditional_token_account.amount as u128
+                * question.payout_numerators[payout_index] as u128)
+                / question.payout_denominator as u128) as u64;
```
0xbigz, line 58: in addition to dust from round downs (perhaps using `div_ceil` for one can mitigate)

doing a mix of account loader `Context` and (unsafe) math logic within one function. would be great to split up math and state modification logic from solana account loader stuff. this makes rust tests much easier to write and lets you fuzz more easily

programs/autocrat/src/instructions/finalize_proposal.rs
metaproph3t: yep, this was indeed a critical issue. resolved: e616e3a83ba43c6fc94cef4d71e53fdbc0658b4f

programs/conditional_vault/src/instructions/initialize_conditional_vault.rs
metaproph3t: that's retarded, okay I guess I will do what anchor does: https://github.com/coral-xyz/anchor/blob/340e9c13f42191a1caa6092d811ed930a64c1f7b/lang/syn/src/codegen/accounts/constraints.rs#L1603-L1627

metaproph3t: fixed: 166de60ab202d30c2f1274b0abbd8cd8dbdf5df8

programs/conditional_vault/src/instructions/redeem_tokens.rs
metaproph3t: yep, I'm okay with that

programs/conditional_vault/src/instructions/redeem_tokens.rs
metaproph3t: makes it easier for me to index stuff if they're not closed

programs/conditional_vault/src/state/conditional_vault.rs
metaproph3t: yep, good catch!

metaproph3t: fixed: https://github.com/metaDAOproject/futarchy/pull/246/commits/963ee252542ce8bc8b825f7a537a0203edd5a60f

programs/conditional_vault/src/instructions/redeem_tokens.rs
metaproph3t: `div_ceil` could cause liabilities to exceed assets as it rounds up. we'd want to use `div_floor`, which is the same as regular division for unsigned integers: https://doc.rust-lang.org/std/primitive.u32.html#method.div_floor

PR #247, opened 2024-08-30
fix(scripts): Update `mintConditionalTokens` to Handle Fractional Units

### LINKED ISSUE #245: mintConditionalTokens doesn't accept fractional units of tokens
https://github.com/metaDAOproject/futarchy/blob/97b19885ee5db7942896ecb06856e5198ad21b08/scripts/main.ts#L756C68-L756C77

https://github.com/metaDAOproject/futarchy/blob/97b19885ee5db7942896ecb06856e5198ad21b08/sdk/src/ConditionalVaultClient.ts#L69

If a user was to pass a fractional amount it gets cast as a big number and rounded to zero if not a whole number, ideally it can be parsed and cast as a BN after calculating the fractional amount.

CC: @inkprojectsol

0xIchigo:
This PR aims to resolve #245 by:
- Converting the user-provided `input` to the smallest unit of the token based on its decimals
- Calculate the scaled amount by multiplying the input amount by 10^(token's decimals)
- Round the scaled amount to the nearest integer
- Create a new `BN` with the scaled, rounded amount

This PR also adds the `getTokenDecimals` helper function to streamline fetching a token's decimals

### DIFF
--- a/sdk/src/ConditionalVaultClient.ts
+++ b/sdk/src/ConditionalVaultClient.ts
@@ -24,8 +24,6 @@ import {
   getAssociatedTokenAddressSync,
 } from "@solana/spl-token";
 
-import { getTokenDecimals } from "../../scripts/main";
-
 export type CreateVaultClientParams = {
   provider: AnchorProvider;
   conditionalVaultProgramId?: PublicKey;
@@ -76,8 +74,7 @@ export class ConditionalVaultClient {
     const storedVault = await this.getVault(vault);
 
     try {
-      const tokenDecimals = await getTokenDecimals(storedVault.underlyingTokenMint);
-      const scaledAmount = uiAmount * Math.pow(10, tokenDecimals);
+      const scaledAmount = uiAmount * Math.pow(10, storedVault.decimals);
       const bnAmount = new BN(scaledAmount.toFixed(0));
 
       return this.mintConditionalTokensIx(
--- a/sdk/src/ConditionalVaultClient.ts
+++ b/sdk/src/ConditionalVaultClient.ts
@@ -24,6 +24,8 @@ import {
   getAssociatedTokenAddressSync,
 } from "@solana/spl-token";
 
+import { getTokenDecimals } from "../../scripts/main";
+
 export type CreateVaultClientParams = {
   provider: AnchorProvider;
   conditionalVaultProgramId?: PublicKey;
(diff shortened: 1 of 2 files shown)


Henry-E:
is this why when someone tries to place an order of 0.69 base tokens it turns into an order for 0.68989898 base tokens?

R-K-H:
> is this why when someone tries to place an order of 0.69 base tokens it turns into an order for 0.68989898 base tokens?

Nah the UI doesn't actually use this, it uses the create IX with the rpc() call and the amount output is actually produced through the simSwap functionality.

R-K-H (approved):
Thanks for doing this!

R-K-H (changes requested):
Actually this happens in another place as well, just something which would wrap up the issue and complete it...

https://github.com/metaDAOproject/futarchy/blob/97b19885ee5db7942896ecb06856e5198ad21b08/sdk/src/ConditionalVaultClient.ts#L80

0xIchigo:
Should be all good now @R-K-H! I opted to just re-use `getTokenDecimals` as a more standard, reusable way of fetching a token's decimals

R-K-H (approved):
LGTM! Thanks for the update provides some higher functionality for the SDK use vs folks diving into the mintConditionalTokensIx() :)

sdk/src/ConditionalVaultClient.ts
```diff
@@ -73,18 +75,20 @@ export class ConditionalVaultClient {
   ) {
     const storedVault = await this.getVault(vault);
 
-    return (
-      this.mintConditionalTokensIx(
+    try {
+      const tokenDecimals = await getTokenDecimals(storedVault.underlyingTokenMint);
```
metaproph3t: you can just use storedVault.decimals for this and don't need to make the extra call

sdk/src/ConditionalVaultClient.ts
```diff
@@ -24,6 +24,8 @@ import {
   getAssociatedTokenAddressSync,
 } from "@solana/spl-token";
 
+import { getTokenDecimals } from "../../scripts/main";
```
metaproph3t: idk if this will work

metaproph3t (changes requested):
Thanks for this! Left 1 comment

PR #249, opened 2024-09-09
feat: try catch for getHumanPrice and cleanup

LukasDeco:


sdk/src/utils/priceMath.ts
```diff
@@ -32,14 +32,19 @@ export class PriceMath {
...
+    );
+    const price1e12 =
       quoteDecimals > baseDecimals
         ? ammPrice.div(decimalScalar)
         : ammPrice.mul(decimalScalar);
-
-    return price1e12.toNumber() / 1e12;
+  
+    try {
+      return price1e12.toNumber() / 1e12;
```
R-K-H, line 45: It'd be good to add a comment in here why the try catch is necessary?

R-K-H (approved):

PR #259, opened 2025-01-14
feat: adds conditional balance with check for swap

R-K-H:
This extends the mint and swap functionality to accept a conditional balance to be passed and either assigns a mint transaction or does not.

It is optional to pass this parameter, but if it does then it will ensure do dynamically mint if necessary.

### DIFF
--- a/sdk/src/v0.3/FutarchyClient.ts
+++ b/sdk/src/v0.3/FutarchyClient.ts
@@ -43,6 +43,7 @@ export class FutarchyClient {
     payer,
     swapType,
     outcome,
+    conditionalAmount,
   }: {
     proposal: PublicKey;
     inputAmount: BN;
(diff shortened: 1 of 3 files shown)


sdk/src/v0.3/FutarchyClient.ts
```diff
@@ -67,16 +69,25 @@ export class FutarchyClient {
...
-        underlyingTokenMint,
-        inputAmount,
-        user,
-        payer
-      )
-      .transaction();
+    // Check to see if we need to mint conditional tokens
+    // If conditionalAmount is not passed we will mint all of the input amount
+    // If conditionalAmount is passed we will mint the difference between the input amount and the conditional amount
+    if (!conditionalAmount || (conditionalAmount && conditionalAmount.lt(inputAmount))) {
```
Dodecahedr0x: can be simplified to `!conditionalAmount || conditionalAmount < inputAmount` because if the first part evaluates to `true` the evaluation stops because it's an `OR`

Dodecahedr0x (commented):
lgtm

sdk/src/v0.3/FutarchyClient.ts
R-K-H: Got it, thanks! I was curious about that given null or undefined.

metaproph3t:
LGTM

PR #269, opened 2025-03-14
Launchpad scripts

pileks:
Added `.env.example` with settings for launches. Move into `.env`, change everything, and then simply hold enter throughout the script. If you don't, you can always enter everything manually.

To initialize a launch, run:
```sh
yarn launch-init
```

To start a launch, run:
```sh
yarn launch-start
```

To complete a launch, run:
```sh
yarn launch-complete
```

### DIFF
--- a/sdk/src/v0.4/LaunchpadClient.ts
+++ b/sdk/src/v0.4/LaunchpadClient.ts
@@ -117,7 +117,8 @@ export class LaunchpadClient {
     secondsForLaunch: number,
     tokenMint: PublicKey,
     launchAuthority: PublicKey = this.provider.publicKey,
-    isDevnet: boolean = false
+    isDevnet: boolean = false,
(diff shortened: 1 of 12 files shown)


sdk/src/v0.4/LaunchpadClient.ts
```diff
@@ -153,10 +154,11 @@ export class LaunchpadClient {
         tokenMint,
         tokenMetadata,
         tokenMetadataProgram: MPL_TOKEN_METADATA_PROGRAM_ID,
+        payer,
       })
       .preInstructions([
         createAssociatedTokenAccountIdempotentInstruction(
-          launchAuthority,
+          payer,
```
pileks, line 161: This is changed because it should be the payer (gas sponsor), not the actual launch authority itself.
Rationale: Launch authority should simply be the one that signs off the launch (init, start), e.g. it could be a non-funded keypair.

PR #274, opened 2025-03-20
Add dao-init and proposal-init scripts in CLI form

pileks:


scripts/daoInit.ts
```diff
@@ -0,0 +1,128 @@
+import { Keypair } from "@solana/web3.js";
```
metaproph3t, line 1: nit: can we call files phrases like 'initDao' instead of 'daoInit'

scripts/daoInit.ts
```diff
@@ -0,0 +1,128 @@
...
+
+import dotenv from "dotenv";
+import * as token from "@solana/spl-token";
+import { BN } from "bn.js";
+
+dotenv.config();
+
+const rpcUrl = await input({
+  message: "Enter your RPC URL:",
+  default: process.env.RPC_URL,
```
metaproph3t, line 17: wdyt about the default devnet and mainnet rpc urls from the setup metric market script?

scripts/daoInit.ts
```diff
@@ -0,0 +1,128 @@
...
+  );
+  console.log("Minted 100,000 META to payer");
+
+  await token.mintTo(
+    provider.connection,
+    payer,
+    usdcMint,
+    usdcAccount,
+    payer,
+    2_000_000n * 1_000_000n // 200,000 USDC with 6 decimals (like in tests)
```
metaproph3t, line 100: comment incorrect

scripts/daoInit.ts
```diff
@@ -0,0 +1,128 @@
...
+    minBaseFutarchicLiquidity,
+    minQuoteFutarchicLiquidity,
+    usdcMint,
+    daoKeypair,
+    twapStartDelaySlots
+  );
+
+  console.log("DAO created with address:", dao.toString());
+  console.log("DAO keypair public key:", daoKeypair.publicKey.toString());
+}
```
metaproph3t, line 122: can we delete the other scripts in the regular anchor style that these are replacing?

scripts/daoInit.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  // Create USDC mint
+  const usdcMint = await token.createMint(
+    provider.connection,
+    payer,
+    payer.publicKey,
+    payer.publicKey,
+    6 // 6 decimals for USDC
+  );
+  console.log("Created USDC mint:", usdcMint.toString());
```
metaproph3t, line 64: it would be kinda cool if the default was to create a new mint, but we could also use this with existing mints, that way we can use the script in prod

PR #320, opened 2025-08-01
Adds a token converter for use with migration and redemption

R-K-H:


programs/token_converter/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+        config.authority = ctx.accounts.authority.key();
+        config.inbound_token_mint = inbound_mint.key();
+        config.outbound_token_mint = outbound_mint.key();
+        config.inbound_token_decimals = inbound_mint.decimals;
+        config.outbound_token_decimals = outbound_mint.decimals;
+        config.max_inbound_token_amount = max_inbound_token_amount;
+        config.max_outbound_token_amount = max_outbound_token_amount;
+        config.burn_inbound_token = burn_inbound_token;
+        config.seq_num = 0;
+        config.bump = ctx.bumps.token_converter_config;
```
metaproph3t: Use `.set_inner` like I do in other programs

programs/token_converter/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+        
+        config.authority = ctx.accounts.authority.key();
+        config.inbound_token_mint = inbound_mint.key();
+        config.outbound_token_mint = outbound_mint.key();
+        config.inbound_token_decimals = inbound_mint.decimals;
+        config.outbound_token_decimals = outbound_mint.decimals;
+        config.max_inbound_token_amount = max_inbound_token_amount;
+        config.max_outbound_token_amount = max_outbound_token_amount;
+        config.burn_inbound_token = burn_inbound_token;
+        config.seq_num = 0;
```
metaproph3t: no event = no seq_num needed

programs/token_converter/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+        let inbound_mint = &ctx.accounts.inbound_token_mint;
+        let outbound_mint = &ctx.accounts.outbound_token_mint;
+        
+        config.authority = ctx.accounts.authority.key();
+        config.inbound_token_mint = inbound_mint.key();
+        config.outbound_token_mint = outbound_mint.key();
+        config.inbound_token_decimals = inbound_mint.decimals;
+        config.outbound_token_decimals = outbound_mint.decimals;
+        config.max_inbound_token_amount = max_inbound_token_amount;
+        config.max_outbound_token_amount = max_outbound_token_amount;
```
metaproph3t: I'd call these supply or something

programs/token_converter/src/lib.rs
```diff
@@ -0,0 +1,310 @@
+use anchor_lang::prelude::*;
+use anchor_spl::token::{self, Token, TokenAccount, Transfer, Burn, Mint};
+use anchor_spl::associated_token::AssociatedToken;
+
+declare_id!("tknMiQZDHrrJe4VDESf3cJorj1jWCfCYK2g4d7nqjT1");
```
metaproph3t: I'd change the structure of this file to be:
- account structs
- derive account strcuts
- instructions
- error

programs/token_converter/src/lib.rs
```diff
@@ -0,0 +1,310 @@
+use anchor_lang::prelude::*;
+use anchor_spl::token::{self, Token, TokenAccount, Transfer, Burn, Mint};
+use anchor_spl::associated_token::AssociatedToken;
```
metaproph3t: add security_txt

programs/token_converter/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+        converter.inbound_token_vault = ctx.accounts.inbound_token_vault.key();
+        converter.outbound_token_vault = ctx.accounts.outbound_token_vault.key();
+        converter.inbound_token_amount = 0;
+        converter.outbound_token_amount = 0;
+        converter.max_inbound_token_amount = config.max_inbound_token_amount;
+        converter.max_outbound_token_amount = config.max_outbound_token_amount;
+        converter.burn_inbound_token = config.burn_inbound_token;
+        converter.inbound_token_decimals = config.inbound_token_decimals;
+        converter.outbound_token_decimals = config.outbound_token_decimals;
+        converter.bump = ctx.bumps.token_converter;
```
metaproph3t: same here `set_inner`

programs/token_converter/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+        config.max_inbound_token_amount = max_inbound_token_amount;
+        config.max_outbound_token_amount = max_outbound_token_amount;
+        config.burn_inbound_token = burn_inbound_token;
+        config.seq_num = 0;
+        config.bump = ctx.bumps.token_converter_config;
+        
+        Ok(())
+    }
+
+    pub fn initialize_token_converter(
```
metaproph3t, line 30: don't think you need separate `config` and `converter`

programs/token_converter/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+
+    pub fn convert(ctx: Context<Convert>, amount: u64) -> Result<()> {
+        let converter = &ctx.accounts.token_converter;
+        let config = &ctx.accounts.token_converter_config;
+        
+        // Validate that amount is greater than 0
+        require!(amount > 0, TokenConverterError::InvalidAmount);
+        
+        // Validate that user has sufficient balance
+        require!(
```
metaproph3t: use `require_gte`

metaproph3t: it logs things out nicer

programs/token_converter/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+            ctx.accounts.from.amount >= amount, 
+            TokenConverterError::InsufficientBalance
+        );
+        
+        // Calculate conversion ratio based on max amounts
+        // This represents how many outbound tokens per inbound token
+        let conversion_ratio = config.max_outbound_token_amount
+            .checked_div(config.max_inbound_token_amount)
+            .ok_or(TokenConverterError::Overflow)?;
```
metaproph3t: what if we just stored this in the config instead of max and max? feels simpler

programs/token_converter/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+        
+        // Calculate amount to transfer
+        // Apply conversion ratio to the original amount
+        let amount_to_transfer = amount
+            .checked_mul(conversion_ratio)
+            .ok_or(TokenConverterError::Overflow)?;
```
metaproph3t: what about reverse splits? I think conversion ratio should be scaled by `1e12` and then you multiply by conversion ratio and divide by 1e12. means you'll need to store these intermediate values in a `u64` instead of `u128` (otherwise it can overflow)

programs/token_converter/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+        
+        // Validate that converter has sufficient outbound tokens
+        require!(
```
metaproph3t: require_gte again

programs/token_converter/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+            ctx.accounts.outbound_token_vault.amount >= amount_to_transfer, 
+            TokenConverterError::InsufficientConverterBalance
```
metaproph3t: this shouldn't happen in practice, right?

metaproph3t: cool to have the invariant check tho

programs/token_converter/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+            // Transfer the inbound tokens from user's account to converter's vault
+            let transfer_in_cpi_accounts = CpiContext::new(
+                ctx.accounts.token_program.to_account_info(),
+                Transfer {
+                    from: ctx.accounts.from.to_account_info(),
+                    to: ctx.accounts.inbound_token_vault.to_account_info(),   
+                    authority: ctx.accounts.authority.to_account_info(),
+                }
+            );
+            token::transfer(transfer_in_cpi_accounts, amount)?;
```
metaproph3t: let's just do transfer for everything

programs/token_converter/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+        }
+
+        // Transfer outbound tokens from converter's vault to user's account
+        let authority_seeds = &[
+            b"token_converter",
+            config.inbound_token_mint.as_ref(),
+            config.outbound_token_mint.as_ref(),
```
metaproph3t: there should be some nonce in here to prevent frontrunning, you could also use the creator key

programs/token_converter/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+            b"token_converter_config",
+            inbound_token_mint.key().as_ref(),
+            outbound_token_mint.key().as_ref()
+        ],
+        bump
+    )]
+    pub token_converter_config: Account<'info, TokenConverterConfig>,
+    
+    pub inbound_token_mint: Account<'info, Mint>,
+    pub outbound_token_mint: Account<'info, Mint>,
```
metaproph3t: check they're not the same

programs/token_converter/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+    
+    pub token_converter_config: Account<'info, TokenConverterConfig>,
+    
+    #[account(mut)]
+    pub authority: Signer<'info>,
+    
+    #[account(
+        mut,
+        constraint = from.mint == token_converter_config.inbound_token_mint @ TokenConverterError::InvalidInboundToken,
+        constraint = from.owner == authority.key() @ TokenConverterError::InvalidAuthority
```
metaproph3t: there are token constraints that do this for you, check some of the other programs

PR #323, opened 2025-08-14
first pass at adding redeem program

jrooks7:


scripts/consts.ts
```diff
@@ -48,3 +48,19 @@ export const DEVNET_DRIFT = new PublicKey(
...
+export const RAYDIUM_CP_SWAP_PROGRAM_ID = new PublicKey(
+  "CPMMoo8L3F4NbTegBCKVNunggL7H1ZpdTHKxQB5qKP1C"
+)
+
+export const MEMO_PROGRAM_ID = new PublicKey(
+  "MemoSq4gqABAXKb96qnH8TysNcWxMyWCqXgDLGmfcHr"
+)
+
+export const TOKEN_PROGRAM_2022_ID = new PublicKey(
+  "TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb"
```
metaproph3t: these should come from SDK

programs/redeem/src/instructions/redeem.rs
```diff
@@ -0,0 +1,273 @@
...
+use anchor_spl::associated_token::AssociatedToken;
+use raydium_cpmm_cpi::{cpi, program::RaydiumCpmm, states::{PoolState, POOL_LP_MINT_SEED}};
+use autocrat::state::Dao;
+use autocrat::program::Autocrat;
+use std::str::FromStr;
+
+use crate::error::RedeemError;
+
+#[derive(Accounts)]
+pub struct UnwindAndMigrate<'info> {
```
metaproph3t: this should have the same name as the file, so either change file or change this

programs/redeem/src/instructions/redeem.rs
```diff
@@ -0,0 +1,273 @@
...
+    // Raydium pool accounts
+    /// The Raydium CPMM pool state (zero-copy account)
+    #[account(
+        mut,
+        seeds = [
+            b"pool_state",
+            dao.key().as_ref(),
+        ],
+        bump,
+        seeds::program = Pubkey::from_str("AfJJJ5UqxhBKoE3grkKAZZsoXDE9kncbMKvqSHGsCNrE").unwrap(), // v4 launchpad
```
metaproph3t: I'd put this in a const

programs/redeem/src/instructions/redeem.rs
```diff
@@ -0,0 +1,273 @@
...
+        );
+        require_keys_eq!(
+            pool.lp_mint,
+            self.lp_mint.key(),
+            RedeemError::InvalidPoolConfigurationLpMint
+        );
+        
+        // Check pool status - ensure withdrawals are enabled
+        require!(
+            pool.status & 2 == 0,
```
metaproph3t: could we use 0b10 instead

programs/redeem/src/instructions/redeem.rs
```diff
@@ -0,0 +1,273 @@
...
+        let cpi_program = ctx.accounts.cp_swap_program.to_account_info();
+        
+        // Execute the withdraw with PDA signer
+        let cpi_ctx = CpiContext::new_with_signer(
+            cpi_program,
+            cpi_accounts,
+            signer_seeds,
+        );
+        
+        cpi::withdraw(
```
metaproph3t: nit: we should have this be `raydium` or `raydium_cpi`

metaproph3t: or even `raydium_cpmm_cpi`

programs/redeem/src/instructions/redeem.rs
```diff
@@ -0,0 +1,273 @@
...
+        constraint = lp_mint.key() == pool_state.load()?.lp_mint @ RedeemError::InvalidPoolConfiguration,
+    )]
+    pub lp_mint: Box<Account<'info, Mint>>,
+    
+    // Mints - validated against DAO and pool configuration
+    /// Base token mint (must match DAO and pool configuration)
+    #[account(
+        mut,
+        constraint = base_mint.key() == dao.base_mint @ RedeemError::InvalidMint,
+        constraint = base_mint.key() == pool_state.load()?.token_0_mint @ RedeemError::InvalidPoolConfiguration,
```
metaproph3t: this'll fail because you need to do this logic to identify `token_0`: https://github.com/metaDAOproject/futarchy/blob/5701e18de9e241a0b3c6cae40f8bb880444b68c2/programs/launchpad/src/instructions/complete_launch.rs#L347-L366

metaproph3t: https://github.com/metaDAOproject/futarchy/blob/5701e18de9e241a0b3c6cae40f8bb880444b68c2/programs/launchpad/src/instructions/complete_launch.rs#L1-L519

programs/redeem/src/instructions/redeem.rs
```diff
@@ -0,0 +1,273 @@
...
+        constraint = quote_mint.key() == dao.quote_mint @ RedeemError::InvalidMint,
+        constraint = quote_mint.key() == pool_state.load()?.token_1_mint @ RedeemError::InvalidPoolConfiguration,
+    )]
+    pub quote_mint: Box<Account<'info, Mint>>,
+    
+    // Source accounts (treasury owns these)
+    /// Treasury's LP token account
+    #[account(
+        mut,
+        constraint = lp_account.owner == treasury.key() @ RedeemError::InvalidAuthority,
```
metaproph3t: I would use the `anchor_spl` constraints here like `associated_token::owner = treasury`

programs/redeem/src/instructions/redeem.rs
```diff
@@ -0,0 +1,273 @@
...
+    )]
+    pub lp_account: Box<Account<'info, TokenAccount>>,
+    
+    /// Treasury's base token account
+    #[account(
+        mut,
+        constraint = base_account.owner == treasury.key() @ RedeemError::InvalidAuthority,
+        constraint = base_account.mint == base_mint.key() @ RedeemError::InvalidTokenAccount,
+    )]
+    pub base_account: Box<Account<'info, TokenAccount>>,
```
metaproph3t: I would rename these `treasury_base_account` and the same for quote

programs/redeem/src/instructions/redeem.rs
```diff
@@ -0,0 +1,273 @@
...
+    pub pool_base_vault: Box<Account<'info, TokenAccount>>,
+    
+    /// Raydium pool's USDC vault, vault 1
+    #[account(
+        mut,
+        constraint = pool_quote_vault.key() == pool_state.load()?.token_1_vault @ RedeemError::InvalidPoolVault,
+        constraint = pool_quote_vault.owner == pool_authority.key() @ RedeemError::InvalidPoolVault,
+        constraint = pool_quote_vault.mint == quote_mint.key() @ RedeemError::InvalidPoolVault,
+    )]
+    pub pool_quote_vault: Box<Account<'info, TokenAccount>>,
```
metaproph3t, line 125: you may be able to just get them as `pool_token_0_vault` and token 1 so that you don't need to do the if else later

programs/redeem/src/instructions/redeem.rs
```diff
@@ -0,0 +1,273 @@
...
+    
+    #[account(
+        mut,
+        constraint = migrator_vault.mint == quote_mint.key() @ RedeemError::InvalidDestination,
+    )]
+    pub migrator_vault: Box<Account<'info, TokenAccount>>,
```
metaproph3t, line 137: we should probably? check that this is actually on a migrator?

metaproph3t: maybe not, but think more about this

scripts/consts.ts
jrooks7: done, MEMO_PROGRAM_ID and TOKEN_PROGRAM_2022_ID don't exist in the SDK so I left them in scripts/consts.ts

PR #324, opened 2025-08-20
Add streamflow test redeem

jrooks7:


programs/redeem/src/instructions/redeem.rs
```diff
@@ -0,0 +1,262 @@
...
+        // seeds = [
+        //     b"vault",
+        //     MIGRATOR_ADMIN_ADDRESS.as_ref(),
+        //     base_mint.key().as_ref(),
+        //     quote_mint.key().as_ref(),
+        // ],
+        // bump,
+        // seeds::program = TOKEN_MIGRATOR_PROGRAM,
+    )]
+    pub migrator_vault: UncheckedAccount<'info>,
```
metaproph3t: this is token account, right? if so we should validate + mint

programs/redeem/src/instructions/redeem.rs
```diff
@@ -0,0 +1,262 @@
...
+            memo_program: ctx.accounts.memo_program.to_account_info(),
+        };
+        
+        let cpi_program = ctx.accounts.cp_swap_program.to_account_info();
+        
+        // Execute the withdraw with PDA signer
+        let cpi_ctx = CpiContext::new(
+            cpi_program,
+            cpi_accounts,
+        );
```
metaproph3t: can we combine this all into one? so no separate `cpi_accounts` and `cpi_program`?

PR #340, opened 2025-09-30
Add spending limit instruction to v0.5 autocrat

metaproph3t:


programs/autocrat/src/instructions/execute_spending_limit_change.rs
R-K-H, line 82: This is not the executor so it will fail the instruction

R-K-H:
This was deployed and only effected one DAO so we'll close this, but keep the branch for now.

PR #343, opened 2025-10-13
Skip token CPIs if amount is 0

ChiefWoods:
Redeem_tokens instruction under conditional_vault program invokes burn and transfer CPIs even if token amount is 0. This change skips the CPI conditionally, shaving off some CUs.


metaproph3t:
Thanks for the PR! I prefer keeping programs as deterministic / simple as possible, and since we're not CU-constrained, I prefer to leave it as-is

ChiefWoods:
Let me know if there's anything else I can help out with

PR #344, opened 2025-10-13
WIP: Recover nested tokens from ATA of ATA

deanmlittle:
WIP fix to enable recovery of ICO funds from an ATA erroneously derived from the vault ATA back to the correct vault address. This may not be suitable to handle all recovery purposes, in which case we may opt to have some kind of a recovery address to recover the funds to in order to refund people who don't understand how token accounts work on Solana.


R-K-H:
Implemented the change and have it integrated. thanks for the support <3

PR #351, opened 2025-10-21
dropped conditional swap eq for proposal count and uodated launch eve…

jrooks7:
…nts to match anchor data


sdk/src/v0.6/FutarchyClient.ts
```diff
@@ -913,6 +913,40 @@ export class FutarchyClient {
...
+    proposal: PublicKey;
+    dao: PublicKey;
+    baseMint: PublicKey;
+    amount: BN;
+    staker?: PublicKey;
+  }) {
+    const stakeAccount = PublicKey.findProgramAddressSync(
+      [Buffer.from("stake"), proposal.toBuffer(), staker.toBuffer()],
+      this.getProgramId(),
+    )[0];
```
metaproph3t: this should be done in a separate function like it's done elsewhere in the SDK (or at least should be done)

PR #366, opened 2025-11-25
Bid wall program

pileks:
# Bid Wall Program

Introduces a new onchain program that allows token holders to sell DAO tokens at the calculated Net Asset Value (NAV) price.

## Motivation

This provides a guaranteed exit mechanism for token holders. Instead of selling on the open market (which may lack liquidity or have unfavorable slippage), holders can always sell directly to the bid wall at fair value derived from the DAO's underlying assets.

### NAV Calculation

The price is computed as `NAV / active_supply`, where:

**NAV** = sum of USDC holdings across:
- DAO treasury (multisig vault)
- Futarchy AMM reserves
- Meteora DAMM pool position (converted from liquidity using the concentrated liquidity curve)

**Active Supply** = total token supply minus tokens held in:
- Futarchy AMM
- Meteora DAMM pool

This ensures DAO-owned tokens are excluded from the denominator, reflecting only circulating supply.

### Selling Flow

1. User calls `sell_tokens` with an amount of base tokens
2. Program computes the USDC output at NAV price
3. A 1% fee (100 bps, rounded up) is deducted
4. User's tokens are **burned** (reducing total supply)
5. USDC is transferred from the bid wall to the user

Token burning is critical: it prevents the bid wall from accumulating tokens and maintains the NAV-per-token ratio for remaining holders.

### Lifecycle

- **Initialize**: Authority creates a bid wall, deposits USDC, and sets a `min_duration` lockup
- **Sell**: Any holder can sell tokens while the wall has USDC liquidity
- **Close**: After `min_duration` elapses, authority can close the wall. Accumulated fees go to a fee recipient, remaining USDC returns to the authority

## Key Security Considerations

- Meteora DAMM pool and position accounts are validated via discriminator checks and owner constraints
- Position-to-pool binding is verified to prevent substituting an unrelated position
- All arithmetic uses checked math with explicit overflow handling
- Fee calculation rounds up to favor the protocol
- Output amount rounds down to favor the protocol
- Only the designated authority can close the bid wall, and only after the minimum duration

## Remaining work

- [ ] Add a `collect_fees` instruction that collects fees into a `fee_wallet` at will, callable only by the `authority`
  - Particularly important for bid walls that are intended to stay open for long periods of time
- [ ] Maybe add optionality for the Meteora DAMMv2 pool/position inclusion
- [ ] Add tests for fail/check cases


programs/bid_wall/src/instructions/initialize_bid_wall.rs
```diff
@@ -0,0 +1,116 @@
...
+    error::BidWallError,
+    meteora_state::{Pool, Position},
+    state::BidWall,
+    usdc_mint,
+};
+
+#[derive(AnchorSerialize, AnchorDeserialize, Clone)]
+pub struct InitializeBidWallArgs {
+    pub amount: u64,
+    pub min_duration: u32,
```
metaproph3t, line 18: should have the units, slots or seconds

programs/bid_wall/src/instructions/close_bid_wall.rs
```diff
@@ -0,0 +1,123 @@
...
+                    b"bid_wall",
+                    ctx.accounts.base_mint.key().as_ref(),
+                    ctx.accounts.authority.key().as_ref(),
+                    &[ctx.accounts.bid_wall.pda_bump],
+                ]],
+            ),
+            ctx.accounts.bid_wall.fees_collected,
+        )?;
+
+        ctx.accounts.bid_wall_usdc_token_account.reload()?;
```
metaproph3t, line 82: why?

programs/bid_wall/src/math/curve.rs
```diff
@@ -0,0 +1,98 @@
+use anchor_lang::prelude::*;
+use ruint::aliases::U256;
```
metaproph3t, line 3: is this coming from meteora? if so you should say so in a comment and put a link

R-K-H:
Screenshot 2025-11-25 at 5 16 23 PM

programs/bid_wall/src/instructions/sell_tokens.rs
```diff
@@ -0,0 +1,216 @@
...
+    pub dao_treasury_usdc_token_account: Account<'info, TokenAccount>,
+
+    /// CHECK: Discriminator checked inside validate
+    #[account(owner = damm_v2_cpi::id())]
+    pub pool: UncheckedAccount<'info>,
+
+    /// CHECK: Discriminator and pool checked inside validate
+    #[account(owner = damm_v2_cpi::id())]
+    pub position: UncheckedAccount<'info>,
+
```
metaproph3t, line 53: why not use the native anchor desrialization? seems better than bytemuck

programs/bid_wall/src/instructions/sell_tokens.rs
```diff
@@ -0,0 +1,216 @@
...
+                            .unwrap(),
+                    )
+                    .unwrap(),
+                spot.quote_reserves
+                    .checked_add(
+                        (pass.quote_reserves + fail.quote_reserves)
+                            .checked_div(2)
+                            .unwrap(),
+                    )
+                    .unwrap(),
```
metaproph3t, line 96: this math is crazy haha, I would take the min

programs/bid_wall/src/instructions/sell_tokens.rs
```diff
@@ -0,0 +1,216 @@
...
+        };
+
+        // DAO Treasury
+        let dao_treasury_usdc = ctx.accounts.dao_treasury_usdc_token_account.amount;
+
+        let dao_nav = dao_treasury_usdc
+            .checked_add(dao_futarchy_amm_usdc)
+            .unwrap()
+            .checked_add(dao_damm_usdc)
+            .unwrap();
```
metaproph3t: we don't need checked_add.unwrap because anchor already does that

programs/bid_wall/src/instructions/sell_tokens.rs
```diff
@@ -0,0 +1,216 @@
...
+        Pool::validate_discriminator(pool_discriminator)?;
+
+        let position_data = self.position.try_borrow_data()?;
+        let position_discriminator = &position_data[..8];
+        Position::validate_discriminator(position_discriminator)?;
+
+        let position: &Position = bytemuck::from_bytes(&position_data[8..]);
+        if position.pool != self.pool.key() {
+            return Err(BidWallError::MeteoraDammPositionPoolMismatch.into());
+        }
```
metaproph3t, line 71: how are we validating that this is the right position / pool? it doesn't look like we are, which I would consider a bad security practice if true

programs/bid_wall/src/instructions/sell_tokens.rs
```diff
@@ -0,0 +1,216 @@
...
+
+        Ok(())
+    }
+
+    pub fn handle(ctx: Context<Self>, args: SellTokensArgs) -> Result<()> {
+        let SellTokensArgs { amount_in } = args;
+
+        let token_total_supply = ctx.accounts.base_mint.supply;
```
metaproph3t: don't think you need this as a separate variable

programs/bid_wall/src/instructions/sell_tokens.rs
metaproph3t: actually I'm not sure if this works either, but whatever the case the methodology here should have clear comments that justify it

programs/bid_wall/src/instructions/sell_tokens.rs
```diff
@@ -0,0 +1,216 @@
...
+            .unwrap();
+        let fee_denominator = 10_000;
+
+        let fee = fee_numerator.checked_div(fee_denominator).unwrap() as u64;
+
+        let fee = if fee_numerator.checked_rem(fee_denominator).unwrap() != 0 {
+            fee.checked_add(1).unwrap()
+        } else {
+            fee
+        };
```
metaproph3t: better way to do this is do 10_000 - fee_rate * before fee / 10_000 and then fee is just amount_out_before_fee minus that

programs/bid_wall/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+    use anchor_lang::prelude::declare_id;
+
+    declare_id!("11111111111111111111111111111111");
+}
+
+#[program]
+pub mod bid_wall {
+    use super::*;
+
+    #[access_control(ctx.accounts.validate(&args))]
```
metaproph3t, line 51: where's fee claiming logic?

programs/bid_wall/src/math/safe_math.rs
```diff
@@ -0,0 +1,108 @@
+use anchor_lang::solana_program::msg;
+use ruint::aliases::U256;
+use std::panic::Location;
```
metaproph3t: again if we really want to pill this in I'd put a link, but I don't think this is necessary

programs/bid_wall/src/meteora_state/pool.rs
```diff
@@ -0,0 +1,239 @@
...
+    pub cliff_fee_numerator: u64,
+    // In fee scheduler first_factor: number_of_period, second_factor: period_frequency, third_factor: reduction_factor
+    // in rate limiter: first_factor: fee_increment_bps, second_factor: max_limiter_duration, max_fee_bps, third_factor: reference_amount
+    pub base_fee_mode: u8,
+    pub padding_0: [u8; 5],
+    pub first_factor: u16,
+    pub second_factor: [u8; 8],
+    pub third_factor: u64,
+    pub padding_1: u64,
+}
```
metaproph3t, line 175: think you can not zero_copy these

programs/bid_wall/src/state/bid_wall.rs
```diff
@@ -0,0 +1,24 @@
...
+pub struct BidWall {
+    /// The PDA bump.
+    pub pda_bump: u8,
+    /// The authority of the bid wall.
+    pub authority: Pubkey,
+    /// The mint of the token being sold into the bid wall.
+    pub base_mint: Pubkey,
+    /// The related DAO.
+    pub dao: Pubkey,
+    /// The DAO's Meteora CPMM pool
```
metaproph3t: DAMM v2 isn't a CPMM, at least how we use it

sdk/package.json
```diff
@@ -14,7 +14,8 @@
     "./v0.3": "./dist/v0.3/index.js",
     "./v0.4": "./dist/v0.4/index.js",
     "./v0.5": "./dist/v0.5/index.js",
-    "./v0.6": "./dist/v0.6/index.js"
+    "./v0.6": "./dist/v0.6/index.js",
```
metaproph3t: ?

programs/bid_wall/src/instructions/close_bid_wall.rs
```diff
@@ -0,0 +1,123 @@
...
+        seeds = [b"bid_wall", base_mint.key().as_ref(), authority.key().as_ref()],
+        bump,
+        has_one = authority
+    )]
+    pub bid_wall: Account<'info, BidWall>,
+
+    #[account(mut)]
+    pub payer: Signer<'info>,
+
+    pub authority: Signer<'info>,
```
metaproph3t, line 21: so only the DAO should be allowed to close it? I don't think that should be the case

programs/bid_wall/src/instructions/close_bid_wall.rs
pileks: We want to pull back all remaining USDC in the ATA to the authority, so that we cleanly close both the Bid Wall and its ATA. Without this, we won't have the correct amount loaded into the ATA's account after we withdraw fees.
Granted, we could just withdraw `bid_wall_usdc_token_account.amount - bid_wall.fees_collected`, but this seemed cleaner and easier to reason about.

LMK if you have a preference for the latter.

programs/bid_wall/src/instructions/sell_tokens.rs
pileks: As you mentioned in a previous comment about linking back to Meteora, here I will also link to the relevant source files that also use zero-copy deserialization (therefore bytemuck). I've mainly kept that structure as it ended up being the simplest way to lift the math we need.

programs/bid_wall/src/instructions/sell_tokens.rs
pileks: My reasoning was that, as we split liquidity in half, we could have assumed that, with a 50/50 chance of fail/pass, the reserve values can be approximated to this.
But as I thought a bit more about it, it might make sense to simply double spot reserves during futarchy mode, as they usually get arbed to a fair market value.
What do you think of this approach?

programs/bid_wall/src/instructions/sell_tokens.rs
pileks: Good catch. Since we're storing the position/pool we're using inside of `bid_wall`, I'll add an `address` constraint here.

programs/bid_wall/src/lib.rs
pileks: Currently inside `close_bid_wall`. I've noted that I'll add another instruction for fee claiming, so I'll be adding that today.

programs/bid_wall/src/instructions/sell_tokens.rs
metaproph3t: thinking through this as I write:

50 USDC / 50 META spot pool = $1
100 pUSDC / 25 pMETA pass pool = $4
25 fUSDC / 100 fMETA fail pool $0.25

DAO treasury is 200 USDC, total token supply is 300 META

based on just spot, DAO treasury is 250 USDC and float is 250 META ($1 NAV per token)
based on spot + pass, DAO treasury is 350 USDC and float is 275 META ($1.27 NAV per token)
based on spot + fail, DAO treasury is 275 USDC and float is 150 META ($1.83 NAV per token)

if we assume the market is efficient, I think the doubling thing works. but pessimistically, at this second the proposal could either pass or fail, so either of these cases could be true.

wait I just realized something, couldn't someone bid up the price in the spot pool and then take out a higher NAV through buybacks?

metaproph3t: originally I envisaged this program as just assuming the initial NAV of ICO price minus any cash that'd gone out, and tbh I think that approach might still make sense

metaproph3t: the most pessimistic assumption is that the initial AMM price is the right one, because any deviations from that price should increase NAV, and I like that assumption

metaproph3t: the hard part of this is that it's possible for them to withdraw liquidity

pileks:
Closing in favor of #367

PR #367, opened 2025-11-28
Bid Wall Program (Simplified)

pileks:
Introduces a bid wall program that allows token holders to sell DAO tokens at a NAV-derived price, receiving quote tokens (USDC) in return.

## Background

This is a simplified implementation compared to the `bid-wall-program` branch, which dynamically computed NAV by reading on-chain state from Futarchy AMM pools, Meteora DAMM positions, and DAO treasury balances. That approach introduced complexity around concentrated liquidity math, cross-program account validation, and potential price manipulation vectors through pool state.

This version derives the base price from AMM reserve ratios set at initialization, then applies a NAV adjustment factor based on treasury balance changes.

### Price Determination

The price calculation has two components:

1. **Base Price Ratio**: Set via `initial_amm_base_reserves` and `initial_amm_quote_reserves`, representing the intended NAV price per token as `quote_reserves / base_reserves`.

2. **NAV Adjustment**: The program stores:
   - `initial_nav`: The total NAV at launch time (passed as argument at initialization)
   - `initial_dao_treasury_quote_amount`: The treasury's quote token balance at initialization (passed as argument)
   - `dao_treasury`: Reference to the DAO treasury account (read at sell time)

At sell time, the current NAV is computed as:

```
current_nav = initial_nav + (current_treasury_quote - initial_treasury_quote)
```

This is an **additive adjustment**: if the treasury gains quote tokens, NAV increases by that amount; if it loses tokens, NAV decreases.

The final payout formula:

```
amount_out = (amount_in × initial_quote_reserves / initial_base_reserves)
             × current_nav 
             / initial_nav
```

**Rationale for additive adjustment**: Using `initial_nav + Δtreasury` instead of a simple `current_treasury / initial_treasury` ratio allows us to track DAO spending, which is usually a pre-approved monthly spending limit, and any spends that occur after a proposal passes. A lot of the DAO's NAV can be locked up inside the bid wall, in which case we'd see a significant drop in the treasury, which would be false information. This a very naive approach, and we would want to consider some kind of bookkeeping program for future use with bid walls.

### Selling Flow

1. User calls `sell_tokens` with a token amount
2. Base quote output is calculated: `amount_in × initial_quote_reserves / initial_base_reserves`
3. Current NAV is computed: `initial_nav + current_treasury_quote - initial_treasury_quote`
4. NAV adjustment is applied: `base_output × current_nav / initial_nav`
5. A 1% fee (100 bps) is deducted from the adjusted output
6. User's tokens are burned (not transferred to the bid wall)
7. Net quote tokens are transferred from the bid wall to the user

Token burning ensures the bid wall doesn't accumulate tokens and maintains consistent per-token value for all sellers.

### Fee Collection

Fees accumulate in the bid wall's quote token balance and are tracked via `fees_collected`. The `collect_fees` instruction allows the designated `fee_recipient` to withdraw accumulated fees at any time. This resets the `fees_collected` counter to zero.

### Lifecycle

- **Initialize**: Creator deposits quote tokens, sets the NAV ratio, provides initial NAV and treasury values, sets fee recipient, authority, and specifies minimum duration
- **Sell**: Any holder can sell while the bid wall has quote tokens and hasn't expired
- **Collect Fees**: Anyone can trigger fee collection; fees always go to the designated `fee_recipient`
- **Cancel**: Authority can close the bid wall at any time, returning quote tokens to authority and fees to recipient
- **Close**: Anyone can close the bid wall after `duration_seconds` elapses; fees go to fee recipient, remaining quote tokens return to authority, rent refunded to payer

## Instructions

| Instruction | Access | Description |
|-------------|--------|-------------|
| `initialize_bid_wall` | Creator signs | Creates bid wall, deposits quote tokens, sets price ratio, initial NAV/treasury values, authority, and duration |
| `sell_tokens` | Anyone | Burns tokens, pays out quote tokens (adjusted for NAV) minus fee |
| `collect_fees` | Anyone (to fee_recipient) | Transfers accumulated fees to fee recipient, resets counter |
| `cancel_bid_wall` | Authority signs | Closes bid wall early, returns funds to authority and fees to recipient |
| `close_bid_wall` | Anyone (after expiry) | Closes accounts, returns funds to authority and fees to recipient |

## State

The `BidWall` account stores:

| Field | Type | Description |
|-------|------|-------------|
| `nonce` | `u64` | Unique identifier allowing multiple bid walls per creator per base mint |
| `created_timestamp` | `i64` | Unix timestamp of bid wall creation |
| `fees_collected` | `u64` | Accumulated fees pending collection |
| `initial_amm_base_reserves` | `u64` | Base token reserve value for price calculation |
| `initial_amm_quote_reserves` | `u64` | Quote token reserve value for price calculation |
| `initial_dao_treasury_quote_amount` | `u64` | Treasury quote token balance at initialization (argument) |
| `initial_nav` | `u64` | Total NAV at launch time (argument) |
| `creator` | `Pubkey` | Signed initialization; used in PDA derivation |
| `authority` | `Pubkey` | Can cancel bid wall; receives remaining quote tokens on close |
| `dao_treasury` | `Pubkey` | DAO treasury address (read at sell time) |
| `base_mint` | `Pubkey` | Token mint being sold into the bid wall |
| `fee_recipient` | `Pubkey` | Receives collected fees |
| `duration_seconds` | `u32` | Minimum duration before permissionless close |
| `pda_bump` | `u8` | PDA bump seed |

### Initialization Arguments

| Argument | Type | Description |
|----------|------|-------------|
| `amount` | `u64` | Quote tokens to deposit into the bid wall |
| `nonce` | `u64` | Unique identifier for PDA derivation; allows multiple bid walls per creator per base mint |
| `initial_amm_base_reserves` | `u64` | Base reserves for price ratio |
| `initial_amm_quote_reserves` | `u64` | Quote reserves for price ratio |
| `initial_nav` | `u64` | Total NAV to use as baseline |
| `initial_dao_treasury_quote_amount` | `u64` | Treasury quote balance to use as baseline |
| `duration_seconds` | `u32` | Minimum duration before close |

Note: `initial_nav` and `initial_dao_treasury_quote_amount` are passed as arguments rather than read on-chain. This design choice:
- Allows flexibility in what counts toward NAV (not just on-chain treasury balance)
- Reduces account dependencies at initialization
- Trusts the creator to provide accurate values (creator is incentivized to be accurate since incorrect values affect their deposited funds)

### Access Control
- `initialize_bid_wall`: Creator must sign — deposits quote tokens and sets authority
- `cancel_bid_wall`: Authority must sign — this is the only way to close early
- `close_bid_wall`: Permissionless after expiry (duration enforced)
- `sell_tokens`: Permissionless while active
- `collect_fees`: Permissionless, but fees always route to stored `fee_recipient`

### Creator vs Authority Separation
The `creator` and `authority` roles are separated:
- **Creator**: Signs the initialization transaction, deposits quote tokens from their account, and is used in PDA derivation
- **Authority**: Can cancel the bid wall early; receives remaining quote tokens on close

This separation allows delegation of cancel/close privileges to a different entity (e.g., a multisig or governance account) without requiring that entity to perform the initialization or hold the deposited funds initially. The creator can set `authority` to themselves if no separation is needed.

### NAV Oracle
- Base price uses stored AMM reserve ratios — no external reads
- NAV adjustment reads the DAO treasury quote token account at sell time
- Attack analysis:
  - **Increasing treasury balance**: Increases payout — costs attacker quote tokens, benefits sellers
  - **Decreasing treasury balance**: Decreases payout — requires treasury access (governance), protects bid wall
  - **Manipulation bounds**: Since adjustment is `initial_nav + Δtreasury`, maximum payout increase is bounded by attacker's capital
  - **Zero treasury**: If treasury goes to 0, `current_nav = initial_nav - initial_treasury_quote`. Payout only goes to zero if `initial_treasury_quote >= initial_nav`

### Initialization Trust Model
- Creator provides `initial_nav` and `initial_dao_treasury_quote_amount` as arguments
- If creator provides incorrect values:
  - Overstated NAV: Sellers get overpaid initially, bid wall drains faster (hurts creator's deposited funds)
  - Understated NAV: Sellers get underpaid, fewer sells occur (hurts sellers)
- Creator is economically incentivized to provide accurate values since they deposit the quote tokens

### Duration Enforcement
- `sell_tokens` is blocked after expiry (`BidWallExpired` error)
- `close_bid_wall` is blocked before expiry (`BidWallNotExpired` error)
- `cancel_bid_wall` bypasses duration but requires authority signature

### Fee Recipient Validation
- Fee recipient is fixed at initialization
- Validated via address constraints on `collect_fees`, `cancel_bid_wall`, and `close_bid_wall`
- Cannot be changed after initialization

### Arithmetic Safety
- All price calculations use `u128` intermediates to prevent overflow
- Division-before-multiplication ordering
- Integer division truncates in favor of the protocol (user receives slightly less)


R-K-H:
We need this to account for the treasury balance minus the spend that they have actually spent.

Eg.

- Raise 1M
- 500k in liquidity
- spend 100k

Deduct 100k from the NAV price (it's actually been withdrawn) for the bid program's price.

programs/bid_wall/src/instructions/cancel_bid_wall.rs
```diff
@@ -0,0 +1,118 @@
...
+    pub base_mint: Account<'info, Mint>,
+
+    #[account(address = usdc_mint::id())]
+    pub quote_mint: Account<'info, Mint>,
+
+    pub token_program: Program<'info, Token>,
+    pub system_program: Program<'info, System>,
+}
+
+// TODO: Theoretically, we could merge the logic of this instruction with the close bid wall instruction.
```
metaproph3t, line 46: I think it was smart to separate them

programs/bid_wall/src/instructions/cancel_bid_wall.rs
```diff
@@ -0,0 +1,118 @@
...
+use anchor_spl::token::{self, Mint, Token, TokenAccount, Transfer};
+
+use crate::{state::BidWall, usdc_mint};
+
+#[event_cpi]
+#[derive(Accounts)]
+pub struct CancelBidWall<'info> {
+    #[account(
+        mut,
+        close=payer,
```
metaproph3t, line 11: did `cargo fmt` not catch this?

programs/bid_wall/src/instructions/cancel_bid_wall.rs
```diff
@@ -0,0 +1,118 @@
...
+        seeds = [b"bid_wall", base_mint.key().as_ref(), authority.key().as_ref()],
```
metaproph3t: I might use a nonce here so that a project can create multiple bid walls

programs/bid_wall/src/instructions/cancel_bid_wall.rs
```diff
@@ -0,0 +1,118 @@
...
+        bump,
+        has_one = authority
+    )]
+    pub bid_wall: Account<'info, BidWall>,
+
+    #[account(mut)]
+    pub payer: Signer<'info>,
+
+    // Authority must sign to prevent unauthorized bid wall cancellation on their behalf
+    #[account(address = bid_wall.authority)]
```
metaproph3t, line 20: I like this practice of having both the `has_one` and `address =` constraint, I might start using this given how bad it can be to miss one

programs/bid_wall/src/instructions/cancel_bid_wall.rs
```diff
@@ -0,0 +1,118 @@
...
+    pub authority: Signer<'info>,
+
+    /// CHECK: used for constraints
+    #[account(address = bid_wall.fee_recipient)]
+    pub fee_recipient: UncheckedAccount<'info>,
+
+    #[account(mut, associated_token::mint = quote_mint, associated_token::authority = bid_wall)]
+    pub bid_wall_usdc_token_account: Account<'info, TokenAccount>,
```
metaproph3t: we should be consistent with either 'usdc' or 'quote' terminology, with my preference being quote terminology

programs/bid_wall/src/instructions/sell_tokens.rs
```diff
@@ -0,0 +1,117 @@
...
+
+        Ok(())
+    }
+
+    pub fn handle(ctx: Context<Self>, args: SellTokensArgs) -> Result<()> {
+        let SellTokensArgs { amount_in } = args;
+
+        let amount_out_before_vault_adjustment: u128 = amount_in as u128
+            * ctx.accounts.bid_wall.initial_amm_quote_reserves as u128
+            / ctx.accounts.bid_wall.initial_amm_base_reserves as u128;
```
metaproph3t, line 68: this still doesn't include any expenses, which is a prerequisite of this contract being deployed

programs/bid_wall/src/instructions/sell_tokens.rs
```diff
@@ -0,0 +1,117 @@
...
+
+        let amount_out_before_fee = (amount_out_before_vault_adjustment
+            * ctx.accounts.dao_treasury_usdc_token_account.amount as u128
+            / ctx.accounts.bid_wall.initial_dao_treasury_quote_amount as u128)
+            as u64;
```
metaproph3t: suppose $1M raise, initial NAV of $0.1

team spends $100k, NAV should be $900k

here your math would be 700 / 800 * 0.1 = 0.0875, so slightly off

programs/bid_wall/src/instructions/cancel_bid_wall.rs
pileks: Agree 100%, it's on the TODO for cleanup

programs/bid_wall/src/instructions/cancel_bid_wall.rs
pileks: I was referring more to the internal `handle` which is essentially the same for both. But yeah, with such a simple piece of code, I think it's okay to keep them entirely separate.

programs/bid_wall/src/instructions/cancel_bid_wall.rs
pileks: nope, it only catches if I leave whitespace after `mut,`, for example.

metaproph3t: can we fix pls

programs/bid_wall/src/instructions/sell_tokens.rs
```diff
@@ -0,0 +1,120 @@
...
+
+        let current_nav = ctx.accounts.bid_wall.initial_nav
+            + ctx.accounts.dao_treasury_quote_token_account.amount
+            - ctx.accounts.bid_wall.initial_dao_treasury_quote_amount;
+
+        let amount_out_before_fee = (amount_out_before_vault_adjustment * current_nav as u128
+            / ctx.accounts.bid_wall.initial_nav as u128) as u64;
+
+        let amount_out_after_fee =
+            ((10_000_u128 - FEE_BPS as u128) * amount_out_before_fee as u128 / 10_000_u128) as u64;
```
metaproph3t, line 78: why not store `quote_in_liquidity` and `base_in_liquidity` when you initialize and then just subtract these from both sides, instead of this math (which seems weird / complex to me)?

`let usdc_in_treasury = ctx.accounts.dao_treasury_quote_token.amount + ctx.accounts.bid_wall.initial_quote_in_lp` (with a comment that this is a pessimistic assumption and works because we're always buying / selling below / above NAV, so the LP operations are accretive)
`let floating_supply = ctx.accounts.base_mint.supply - ctx.accounts.bid_wall.initial_base_in_lp` (with similar comment here)
`let amount_out_before_fee = amount_in as u128 * usdc_in_treasury / floating_supply`

I would also be verbose with comments here. this math is extremely important (it will determine how much money entrepreneurs have to run their businesses / how much investors are able to get out in bad times) and we should have solid reasoning for why it's the way it is

pileks: Correct, I was reviewing all of this math last night and came to the same conclusion. I just hit the re-review button too early. 😏 

We'll do that exactly.

pileks: Now that I sat down to solve this, I'm having trouble with the math here.

The problem I see with your assumption is that it doesn't take into account the fact that the amount of quote in the bid wall is technically also quote in the treasury, as it's the project's money (or will be at one point), thus it affects NAV. Otherwise, what is the point of locking up those funds?

Same thing goes for `floating_supply` - it's more than just `mint.supply - initial_base_in_lp`, as we have carveouts and performance packages, although we can assume that those simply reduce NAV.

I'm not saying my math was correct - it was weird, and I don't think it's good. But, even by recording initial base/quote in LP, we don't get a correct pessimistic assumption of the project's NAV, or NAV per token (e.g. the price we want to buy tokens at).

I'm thinking that the correct price would be:

```
dao.treasury_balance + initial_lp_quote + bid_wall_quote
--------------------------------------------------------
           mint.supply - initial_base_in_lp
```

Where `base_in_lp` assumes both Futarchy AMM and Meteora.

Given that our launch price is `final_raise_amount/TOKENS_TO_PARTICIPANTS`, my assumption is that carveouts and performance packages are not included as part of the initial NAV per token, as we're actually looking at NAV per floating token.

With this in mind, I'd change the above formula to:

```
dao.treasury_balance + initial_lp_quote + bid_wall_quote
--------------------------------------------------------
    mint.supply - initial_base_in_lp - locked_tokens
```

Where we would also provide `locked_tokens` to the equation.

Does this make sense to you?

pileks:
closing in favor of #373

PR #387, opened 2025-12-24
Add `return_funds` instruction to Launchpad v6

pileks:
Adds an admin-only instruction that allows the MetaDAO operational multisig to recover USDC from a launch's quote vault by transferring it to a specified recipient.

## Motivation

This instruction serves as a recovery mechanism for funds that may have been mistakenly sent directly to a launch's quote vault outside of normal program flows (e.g., accidental transfers). Without this, such funds would be permanently locked since only the launch's PDA signer can authorize transfers from the vault.

## Changes

- New `return_funds` instruction that transfers a specified amount of USDC from `launch_quote_vault` to a recipient's associated token account
  - Callable only by the MetaDAO operational multisig in production
- New `LaunchFundsReturnedEvent` emitted on each recovery for off-chain tracking
- New `InvalidAdmin` error code

## Notes

- The instruction can be called regardless of launch state (live, completed, failed, etc.) - this is intentional since recovery may be needed at any point
- The admin can specify any recipient address and any amount (up to vault balance)
- No state changes occur on the `Launch` account other than incrementing `seq_num`

### DIFF
--- a/programs/v06_launchpad/src/instructions/return_funds.rs
+++ b/programs/v06_launchpad/src/instructions/return_funds.rs
@@ -1,0 +1,96 @@
+use anchor_lang::prelude::*;
+use anchor_spl::token::{self, Token, TokenAccount, Transfer};
+
+use crate::events::{CommonFields, LaunchFundsReturnedEvent};
+use crate::state::Launch;
+
+pub mod admin {
+    use anchor_lang::prelude::declare_id;
+
+    // MetaDAO multisig vault
+    declare_id!("6awyHMshBGVjJ3ozdSJdyyDE1CTAXUwrpNMaRGMsb4sf");
+}
+
+#[derive(AnchorSerialize, AnchorDeserialize, Clone)]
+pub struct ReturnFundsArgs {
+    pub amount: u64,
+}
+
+#[event_cpi]
+#[derive(Accounts)]
+pub struct ReturnFunds<'info> {
+    pub admin: Signer<'info>,
+
+    #[account(
+        mut,
+        has_one = launch_quote_vault,
+        has_one = launch_signer,
+    )]
+    pub launch: Account<'info, Launch>,
+
+    #[account(mut)]
+    pub launch_quote_vault: Account<'info, TokenAccount>,
+
+    /// CHECK: This is the launch signer
+    #[account(
+        seeds = [b"launch_signer", launch.key().as_ref()],
+        bump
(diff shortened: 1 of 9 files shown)


tests/launchpad/unit/returnFunds.test.ts
```diff
@@ -0,0 +1,115 @@
...
+    );
+
+    const recipientQuoteAccount = getAssociatedTokenAddressSync(
+      MAINNET_USDC,
+      this.payer.publicKey,
+      true,
+    );
+
+    // Transfer wrongly sent USDC to the launch signer
+    await this.transfer(
```
R-K-H, line 85: 1) Wut?

tests/launchpad/unit/returnFunds.test.ts
```diff
@@ -0,0 +1,115 @@
...
+      MAINNET_USDC,
+      this.payer,
+      launchSigner,
+      amount_wrongly_sent.toNumber(),
+    );
+
+    await launchpadClient.launchpad.methods
```
R-K-H, line 92: Should add an idempotent I think...

programs/v06_launchpad/src/instructions/return_funds.rs
```diff
@@ -0,0 +1,96 @@
...
+
+        let seeds = &[
+            b"launch_signer",
+            launch_key.as_ref(),
+            &[launch.launch_signer_pda_bump],
+        ];
+        let signer = &[&seeds[..]];
+
+        // Transfer USDC back to the recipient
+        token::transfer(
```
R-K-H, line 71: Does transfer handle the idempotent create?

pileks: It doesn't (or at least shouldn't).

tests/launchpad/unit/returnFunds.test.ts
pileks: Sure thing. Didn't really think about it as USDC is so ubiquitous, but it makes sense.

tests/launchpad/unit/returnFunds.test.ts
pileks: Launch signer controls the launch USDC vault, so this essentially raw sends USDC to the signer's vault. I've updated the comment to better reflect this.

R-K-H (approved):

PR #389, opened 2026-01-02
Add `burn_performance_package` instruction

pileks:
Adds an admin-only instruction to permanently close performance packages and burn any remaining tokens in their vaults.

## Changes

- New `burn_performance_package` instruction in the `price_based_performance_package` program
- SDK type updates for v0.6 and v0.7
- Squads multisig execution scripts
- Unit tests

## Mechanics

The instruction performs the following operations in sequence:

1. **Burns vault tokens** - If the performance package token vault holds any tokens, they are burned via CPI to the token program
2. **Closes token vault** - The associated token account is closed, returning its rent-exempt lamports to the `spill_account`
3. **Closes performance package** - The account is closed via Anchor's `close` constraint, returning lamports to the `spill_account`

## Access Control

In production builds (compiled with `--features production`), only the MetaDAO operational multisig (`6awyHMshBGVjJ3ozdSJdyyDE1CTAXUwrpNMaRGMsb4sf`) can invoke this instruction. The check is bypassed in non-production builds to enable testing.


programs/price_based_performance_package/src/instructions/burn_performance_package.rs
```diff
@@ -0,0 +1,102 @@
...
+            token::CloseAccount {
+                account: ctx
+                    .accounts
+                    .performance_package_token_vault
+                    .to_account_info(),
+                destination: ctx.accounts.spill_account.to_account_info(),
+                authority: ctx.accounts.performance_package.to_account_info(),
+            },
+            signer,
+        ))?;
```
metaproph3t: I'd rather just remove this I think, marginal code with its resulting surface isn't worth the rent rebate (unless there's another reason)?

pileks: No reason really, just cleanup of state on the blockchain.
Can't do this directly through a `close` Anchor constraint, and this account will never be used again (as new PP will have new token account tied to it).

PR #391, opened 2026-01-06
changing collect fees to model collect meteora fees

jrooks7:


package.json
```diff
@@ -68,7 +68,7 @@
     "ts-mocha": "^10.0.0",
     "ts-node": "^10.9.2",
     "tsx": "^4.7.1",
-    "typescript": "^4.3.5",
+    "typescript": "^5.5.0",
```
pileks: Why the typescript upgrade?

scripts/v0.7/collectAllFees.ts
```diff
@@ -0,0 +1,180 @@
...
+        );
+
+      const signature = await futarchy
+        .collectFeesIx({
+          dao: dao.publicKey,
+          baseMint: dao.account.baseMint,
+          quoteMint: dao.account.quoteMint,
+        })
+        .preInstructions([createBaseAtaIx, createQuoteAtaIx])
+        .rpc();
```
pileks: You should consider adding the base/quote ATA idempotent IXs to both transactions, as the first transaction might fail, and then the base/quote ATAs won't be created for the 2nd transaction.

pileks (changes requested):
I see some tests are failing, look into that please.

Also, review the tests for fee collection and whether they properly check pre/post token balances of the metadao multisig vault, now that fee collection is tied to it.

scripts/v0.7/collectAllFees.ts
jrooks7: Good call out, will fix

tests/futarchy/unit/finalizeProposal.test.ts
```diff
@@ -225,6 +225,13 @@ export default function suite() {
     const storedProposal = await this.futarchy.getProposal(proposal);
     assert.exists(storedProposal.state.passed);
 
+    // Create ATAs for the metadao multisig vault (hardcoded fee destination)
+    const METADAO_MULTISIG = new PublicKey(
+      "6awyHMshBGVjJ3ozdSJdyyDE1CTAXUwrpNMaRGMsb4sf",
+    );
+    await this.createTokenAccount(META, METADAO_MULTISIG);
+    await this.createTokenAccount(USDC, METADAO_MULTISIG);
```
pileks, line 233: This is a good idea to do when we finalize the proposals in our scripts too. At least it seems that way to me.

programs/futarchy/src/instructions/collect_fees.rs
```diff
@@ -1,21 +1,27 @@
 use super::*;
 
-pub mod admin {
+pub mod metadao_multisig_vault {
     use anchor_lang::prelude::declare_id;
 
-    // MetaDAO multisig
+    // MetaDAO multisig - hardcoded fee destination
```
pileks: I always had a problem with the "multisig" naming here. I think that whenever we're talking about multisigs we have to make sure that we're clearly stating whether something is a "multisig" or a "multisig vault". I'd just add "vault" to the comment as well, so that we know from reading this comment where exactly fees are going.

pileks (approved):
Looks good overall! GG

PR #394, opened 2026-01-08
optimize points-based approval script

pileks:


scripts/v0.7/pointsBased/approveWithPointsWeightedPhase.ts
```diff
@@ -29,25 +30,32 @@ const launchpad: LaunchpadClient = LaunchpadClient.createClient({ provider });
...
 async function main() {
-  const pointsAllocations: PointsAllocation[] = JSON.parse(
+  // Script expects a keyed JSON file like:
+  // { "11111111111111111111111111111111": { "points": 100 }, "22222222222222222222222222222222": { "points": 100 } }
+  const pointsAllocations = JSON.parse(
     fs.readFileSync(
       path.join(process.cwd(), "scripts/v0.7/pointsBased/points.json"),
       "utf8",
     ),
-  ).map((x) => ({
```
R-K-H, line 43: I think if you build a set it's even better cause there can be no duplicates in a set.

PR #395, opened 2026-01-08
modifying collect fees on bid wall for accounting purposes

jrooks7:


programs/bid_wall/src/instructions/collect_fees.rs
```diff
@@ -32,6 +43,17 @@ pub struct CollectFees<'info> {
...
+        {
+            use crate::error::BidWallError;
+
+            require_keys_eq!(
+                self.admin.key(),
+                metadao_admin::ID,
+                BidWallError::InvalidAdmin
+            );
+        }
+
```
pileks, line 54: I'd normally do this by just straight saying `crate::error::BidWallError:InvalidAdmin` in the `require_keys_eq`, but this is okay too. If you don't feel strongly eitherway, that's going to produce a couple of lines less.

pileks (approved):
Address the comment according to your wishes, otherwise it's a LGTM 🍏

Please cross-check that we're setting the same fee recipient inside `launchpad`, just to be sure.

programs/bid_wall/src/instructions/collect_fees.rs
jrooks7: less lines the better, will make this change

programs/bid_wall/src/instructions/collect_fees.rs
```diff
@@ -32,6 +43,13 @@ pub struct CollectFees<'info> {
 
 impl CollectFees<'_> {
     pub fn validate(&self) -> Result<()> {
+        #[cfg(feature = "production")]
+        require_keys_eq!(
+            self.admin.key(),
+            metadao_admin::ID,
+            crate::error::BidWallError::InvalidAdmin
```
metaproph3t: should be just `BidWallError::InvalidAdmin`

programs/bid_wall/src/instructions/collect_fees.rs
```diff
@@ -7,20 +7,31 @@ use crate::{
...
 };
 
+pub mod metadao_multisig_vault {
+    use anchor_lang::prelude::declare_id;
+
+    // MetaDAO operations multisig vault - hardcoded fee destination
+    declare_id!("6awyHMshBGVjJ3ozdSJdyyDE1CTAXUwrpNMaRGMsb4sf");
+}
+
+pub mod metadao_admin {
```
metaproph3t: I'd call this something different like `metadao_cranker` to imply that this wallet shouldn't be trusted

programs/bid_wall/src/instructions/collect_fees.rs
```diff
@@ -1,26 +1,39 @@
...
+    declare_id!("tSTp6B6kE9o6ZaTmHm2ZwnJBBtgd3x112tapxFhmBEQ");
+}
+
 #[event_cpi]
 #[derive(Accounts)]
 pub struct CollectFees<'info> {
     #[account(mut)]
     pub bid_wall: Account<'info, BidWall>,
 
+    pub admin: Signer<'info>,
```
metaproph3t: this should be renamed as well

programs/bid_wall/src/error.rs
```diff
@@ -14,4 +14,6 @@ pub enum BidWallError {
     BidWallDepleted,
     #[msg("Invalid input amount")]
     InvalidInputAmount,
+    #[msg("Invalid admin")]
+    InvalidAdmin,
```
metaproph3t: this should be renamed

metaproph3t: maybe fee collector instead of admin, or cranker if we want to be consistent

PR #396, opened 2026-01-10
Optimistic governance

pileks:
This PR introduces optimistic governance, an alternative governance mode that allows the designated team address to propose treasury spends that automatically pass after a time delay unless challenged.

## Overview

Optimistic governance enables faster treasury operations for routine spends while preserving the ability for token holders to challenge and force a full futarchy vote when needed.

### Initiating an Optimistic Proposal

The team address (stored in `dao.team_address`) can call `initiate_vault_spend_optimistic_proposal` to create a token transfer proposal. This instruction:

1. Requires `is_optimistic_governance_enabled` to be `true` on the DAO
2. Requires the DAO to have no active proposal (pool must be in spot state)
3. Requires no other optimistic proposal to be currently enqueued
4. Caps the transfer amount at 3x the DAO's spending limit
5. Creates a Squads vault transaction and proposal
6. Stores the proposal reference and timestamp in `dao.optimistic_proposal`

### Challenge Window

Once enqueued, the optimistic proposal enters a challenge window equal to `seconds_per_proposal`. During this period, anyone can challenge the proposal by initializing, staking to, and launching a futarchy proposal with the same Squads proposal, which:

1. Clears `dao.optimistic_proposal`
2. Launches a standard futarchy proposal
3. The spend must now pass the decision market to be approved (without the benefit of the team-sponsored pass threshold)

### Finalization

After `seconds_per_proposal` elapses without a challenge, anyone can call `finalize_optimistic_proposal` to:

1. Have the DAO approve the Squads proposal
2. Clear `dao.optimistic_proposal`
3. The Squads vault transaction can then be executed normally

## Safety Checks

- **Time-based protection**: Proposals cannot be finalized before `seconds_per_proposal` elapses
- **Challenge cutoff**: Once the time window passes, the proposal cannot be challenged via `initialize_proposal` (returns `OptimisticProposalAlreadyPassed`)
- **Launch guard**: `launch_proposal` also checks that the optimistic proposal hasn't already passed, preventing edge cases where initialization succeeds but launch fails
- **Single proposal**: Only one optimistic proposal can be active at any time
- **Amount cap**: Transfer amounts are bounded by 3x the spending limit
- **No team sponsorship of challenges**: The team cannot sponsor a futarchy proposal that challenges their own optimistic proposal. This prevents the team from benefiting from the `team_sponsored_pass_threshold_bps` when a challenge goes to a decision market.

## State Changes

The `Dao` account gains two new fields:
- `optimistic_proposal: Option<OptimisticProposal>` - tracks the active optimistic proposal if any
- `is_optimistic_governance_enabled: bool` - feature flag (defaults to `false`)

`OptimisticProposal` contains:
- `squads_proposal: Pubkey` - the Squads proposal that will be approved on finalization
- `enqueued_timestamp: i64` - when the optimistic proposal was created

## New Instructions

- `initiate_vault_spend_optimistic_proposal` - team-only, creates a transfer proposal
- `finalize_optimistic_proposal` - permissionless, approves the Squads proposal after time delay

## Refactoring

The `compile_transaction_message` helper was extracted to `src/squads.rs` as `compile_squads_transaction_message` for reuse across instructions that create Squads vault transactions.


programs/futarchy/src/instructions/collect_meteora_damm_fees.rs
```diff
@@ -399,105 +398,3 @@ impl CollectMeteoraDammFees<'_> {
         Ok(())
     }
 }
-
```
R-K-H, line 402: Why is this touched?

programs/futarchy/src/instructions/mod.rs
```diff
@@ -20,9 +22,11 @@ pub use collect_fees::*;
 pub use collect_meteora_damm_fees::*;
```
R-K-H, line 31: If we remove it, then not sure we can import it.

scripts/v0.6/launchSOLO.ts
```diff
@@ -74,8 +74,7 @@ export const launch = async () => {
     .initializeLaunchIx({
       tokenName: "Solomon",
       tokenSymbol: "SOLO",
-      tokenUri:
-        "https://solomonlabs.org/assets/solo.json",
+      tokenUri: "https://solomonlabs.org/assets/solo.json",
```
R-K-H, line 77: again touching things the pr shouldn't or if it should just wanted to know

R-K-H (dismissed):
Only other thing is I don't think another proposal should be able to be created while there exists an unexecuted OGP.

scripts/v0.6/launchSOLO.ts
pileks: This was auto-touched by our linter, so I left it in... :)

programs/futarchy/src/instructions/mod.rs
pileks: Not sure what you're referring to, nothing is removed here...

programs/futarchy/src/instructions/collect_meteora_damm_fees.rs
pileks: Both this and the new IXes use this function to create Squads transactions. Made sense to refactor it into a shared file.

R-K-H: Roger, just wanted to make sure.

programs/futarchy/src/instructions/mod.rs
R-K-H: I just thought your removed the function I didn't see it elsewhere.

programs/futarchy/src/instructions/finalize_optimistic_proposal.rs
```diff
@@ -0,0 +1,87 @@
...
+        );
+
+        // Pool must be in spot state - no active proposals
+        // This should never be hit, but it's here for completeness
+        match self.dao.amm.state {
+            PoolState::Spot { spot: _ } => {}
+            _ => {
+                return Err(FutarchyError::PoolNotInSpotState.into());
+            }
+        }
```
metaproph3t: same comment with the `matches!`

programs/futarchy/src/instructions/initiate_vault_spend_optimistic_proposal.rs
```diff
@@ -0,0 +1,212 @@
...
+pub struct InitiateVaultSpendOptimisticProposal<'info> {
+    #[account(mut, seeds = [squads_multisig_program::SEED_PREFIX, squads_multisig_program::SEED_MULTISIG, dao.key().as_ref()], bump, seeds::program = squads_program)]
+    pub squads_multisig: Account<'info, squads_multisig_program::Multisig>,
+
+    /// CHECK: The squads multisig vault that executes the transaction
+    #[account(seeds = [squads_multisig_program::SEED_PREFIX, squads_multisig.key().as_ref(), squads_multisig_program::SEED_VAULT, 0_u8.to_le_bytes().as_ref()], bump, seeds::program = squads_program)]
+    pub squads_multisig_vault: UncheckedAccount<'info>,
+
+    #[account(seeds = [squads_multisig_program::SEED_PREFIX, squads_multisig.key().as_ref(), squads_multisig_program::SEED_SPENDING_LIMIT, dao.key().as_ref()], bump, seeds::program = squads_program)]
+    pub squads_spending_limit: Account<'info, squads_multisig_program::SpendingLimit>,
```
metaproph3t, line 19: make sure this is always one account

+ add a test for updating spending limit that optimistic governance still works

programs/futarchy/src/instructions/initialize_dao.rs
```diff
@@ -212,6 +212,8 @@ impl InitializeDao<'_> {
             },
             team_sponsored_pass_threshold_bps,
             team_address,
+            optimistic_proposal: None,
+            is_optimistic_governance_enabled: false,
```
metaproph3t, line 225: don't we need to add some update account instructions again here

metaproph3t: theoretically this could be a separate PR but I prefer it in this one

programs/futarchy/src/instructions/initialize_proposal.rs
```diff
@@ -36,6 +36,22 @@ pub struct InitializeProposal<'info> {
 
 impl InitializeProposal<'_> {
     pub fn validate(&self) -> Result<()> {
+        // If we're trying to challenge an optimistic proposal that has already passed due to age, we should error
+        // This is because the optimistic proposal's Squads proposal will eventually have to be executed
+        match self.dao.optimistic_proposal {
```
metaproph3t: I think an if let would be better here

programs/futarchy/src/instructions/initiate_vault_spend_optimistic_proposal.rs
```diff
@@ -0,0 +1,212 @@
...
+
+#[derive(Accounts)]
+#[event_cpi]
+pub struct InitiateVaultSpendOptimisticProposal<'info> {
+    #[account(mut, seeds = [squads_multisig_program::SEED_PREFIX, squads_multisig_program::SEED_MULTISIG, dao.key().as_ref()], bump, seeds::program = squads_program)]
+    pub squads_multisig: Account<'info, squads_multisig_program::Multisig>,
+
+    /// CHECK: The squads multisig vault that executes the transaction
+    #[account(seeds = [squads_multisig_program::SEED_PREFIX, squads_multisig.key().as_ref(), squads_multisig_program::SEED_VAULT, 0_u8.to_le_bytes().as_ref()], bump, seeds::program = squads_program)]
+    pub squads_multisig_vault: UncheckedAccount<'info>,
```
metaproph3t, line 16: wouldn't have this be unchecked unless necessary

programs/futarchy/src/instructions/initialize_dao.rs
pileks: `update_dao` handles updating already. We decided to go with `false` as a default on creation as we don't want it on when a new DAO gets created. We can add a param to creation if you want to.

programs/futarchy/src/instructions/initiate_vault_spend_optimistic_proposal.rs
```diff
@@ -0,0 +1,212 @@
...
+            FutarchyError::OptimisticGovernanceDisabled
+        );
+
+        // Pool must be in spot state - no active proposal
+        match self.dao.amm.state {
+            PoolState::Spot { spot: _ } => {}
+            _ => {
+                return Err(FutarchyError::PoolNotInSpotState.into());
+            }
+        }
```
metaproph3t: matches!

programs/futarchy/src/instructions/initiate_vault_spend_optimistic_proposal.rs
```diff
@@ -0,0 +1,212 @@
...
+        match self.dao.optimistic_proposal {
+            Some(ref optimistic_proposal) => {
+                require_gte!(
+                    Clock::get()?.unix_timestamp,
+                    optimistic_proposal.enqueued_timestamp + self.dao.seconds_per_proposal as i64,
+                    FutarchyError::ProposalDurationTooShort
+                );
+            }
+            None => {}
+        };
```
metaproph3t: I don't like ever enqueuing an optimistic proposal while another one is still there, I'd rather add an instruction to remove the optimistic proposal

programs/futarchy/src/instructions/initiate_vault_spend_optimistic_proposal.rs
```diff
@@ -0,0 +1,212 @@
...
+
+        // Amount must be less than or equal to 3 times the spending limit
+        require_gte!(
+            self.squads_spending_limit.amount.checked_mul(3).unwrap(),
```
metaproph3t: checked_mul unwrap is equivalent to just doing multiplication because anchor has the overflow checks

programs/futarchy/src/instructions/launch_proposal.rs
```diff
@@ -51,6 +51,23 @@ impl LaunchProposal<'_> {
...
+        // If there is an active optimistic proposal, it must be for the same squads proposal
+        // as the futarchy proposal we're launching, thus challenging the optimistic proposal.
+        // This follows the logic that a DAO can have only one proposal active at a time.
+        if let Some(optimistic_proposal) = &self.dao.optimistic_proposal {
+            require_keys_eq!(
+                optimistic_proposal.squads_proposal,
+                self.proposal.squads_proposal
+            );
+
+            // The optimistic proposal must be younger than seconds_per_proposal, otherwise it is considered passed and must be finalized
```
metaproph3t, line 69: also check that it's unpassed

programs/futarchy/src/instructions/launch_proposal.rs
```diff
@@ -139,6 +156,14 @@ impl LaunchProposal<'_> {
         proposal.state = ProposalState::Pending;
         proposal.timestamp_enqueued = clock.unix_timestamp;
 
+        // Update the DAO state
+        if dao.optimistic_proposal.is_some() {
```
metaproph3t, line 192: you don't need to do this, can just do 
`dao.optimistic_proposal = None` and clear out any active proposal

programs/futarchy/src/instructions/sponsor_proposal.rs
```diff
@@ -23,6 +23,15 @@ impl SponsorProposal<'_> {
...
         );
 
+        // Team cannot sponsor a challenge to an optimistic proposal
+        if let Some(optimistic_proposal) = &self.dao.optimistic_proposal {
+            require_keys_neq!(
+                optimistic_proposal.squads_proposal,
+                self.proposal.squads_proposal,
+                FutarchyError::CannotSponsorOptimisticProposalChallenge
+            );
+        }
```
metaproph3t: idk if we need this

programs/futarchy/src/instructions/initiate_vault_spend_optimistic_proposal.rs
pileks: Note - Squads Multisig Vaults don't have any data, they are just derived addresses.
Disregarding this comment for now.

pileks:
After talking to @metaproph3t we came to the following conclusion:

Converting an optimistic proposal to a regular market proposal (by means of using `launch_proposal`) should make that proposal team-sponsored.

programs/futarchy/src/instructions/launch_proposal.rs
pileks: This actually applies in general, so the check will be outside this condition.

programs/futarchy/src/instructions/sponsor_proposal.rs
pileks: After our conversation, we don't.

metaproph3t:
LGTM aside from the account resizing

programs/futarchy/src/instructions/initiate_vault_spend_optimistic_proposal.rs
pileks: https://github.com/metaDAOproject/programs/pull/396/changes#diff-0df48bb71b76c7fddbb1c5391964aa9702a5bc66559bca9f0ad59e623efad1f3R136

R-K-H:
Is this good to merge now?

pileks:
> Is this good to merge now?

@R-K-H No, not until we explicitly say "we want to turn this on".

Prerequisites:
- Add team addresses to existing teams that want optimistic governance (admin update)
- UI for optimistic proposals - extremely important to surface UI to tokenholders

metanallok (approved):

PR #397, opened 2026-01-12
Address audit

pileks:


programs/conditional_vault/src/instructions/common.rs
```diff
@@ -34,12 +34,13 @@ impl<'info, 'c: 'info> InteractWithVault<'info> {
             VaultError::InvalidConditionals
         );
 
-        let mut conditional_token_mints = vec![];
-        let mut user_conditional_token_accounts = vec![];
+        let mut conditional_token_mints = Vec::with_capacity(expected_num_conditional_tokens);
+        let mut user_conditional_token_accounts =
+            Vec::with_capacity(expected_num_conditional_tokens);
```
metaproph3t: let's not do this, the loss of readability isn't worth the CU gains

programs/conditional_vault/src/instructions/initialize_conditional_vault.rs
```diff
@@ -46,17 +50,17 @@ impl<'info, 'c: 'info> InitializeConditionalVault<'info> {
         let remaining_accs = &mut ctx.remaining_accounts.iter();
 
         let expected_num_conditional_tokens = ctx.accounts.question.num_outcomes();
-        let mut conditional_token_mints = vec![];
+        let mut conditional_token_mints = Vec::with_capacity(expected_num_conditional_tokens);
```
metaproph3t: same here

programs/conditional_vault/src/instructions/merge_tokens.rs
```diff
@@ -103,11 +103,9 @@ impl<'info, 'c: 'info> InteractWithVault<'info> {
 
         ctx.accounts.vault.seq_num += 1;
 
+        let clock = Clock::get()?;
         emit_cpi!(MergeTokensEvent {
-            common: CommonFields {
-                slot: Clock::get()?.slot,
-                unix_timestamp: Clock::get()?.unix_timestamp,
-            },
```
metaproph3t, line 110: wow this was really bad code before lol

programs/conditional_vault/src/error.rs
```diff
@@ -36,4 +36,8 @@ pub enum VaultError {
     ConditionalTokenMetadataAlreadySet,
     #[msg("Conditional token account is not owned by the authority")]
     UnauthorizedConditionalTokenAccount,
+    #[msg("Invalid payout numerators")]
```
metaproph3t: would rename comment to "Payout numerators are too large, causing an overflow"

metaproph3t:
left comments, otherwise LGTM

PR #398, opened 2026-01-13
Backdoor execution of multisig proposals

pileks:
Adds an admin backdoor allowing the MetaDAO multisig to directly execute Squads config transactions on DAO multisigs without going through futarchy voting.

### Admin Backdoor for Multisig Config Transactions

New instruction: `admin_approve_execute_multisig_proposal`

DAOs using Squads multisig occasionally need to execute config transactions (updating time locks, changing thresholds, etc.) as part of a DAO's regular proposal process. Sometimed DAOs need to have both vault and config transactions as part of the same proposal, which Squads doesn't support. This instruction allows the MetaDAO multisig to approve and execute such transactions directly, when necessary, and approved by the DAO.

**Constraints:**
- Caller must be the MetaDAO multisig vault (`6awyHMshBGVjJ3ozdSJdyyDE1CTAXUwrpNMaRGMsb4sf`) in production
- DAO's AMM must be in Spot state (no active markets)

### Bonus: Stale Transaction Protection

While implementing this, a potential brick scenario was discovered. When a Squads config transaction executes, it increments `stale_transaction_index`, invalidating all vault transactions created before it. A futarchy proposal pointing to a staled Squads transaction would pass its vote but fail to execute, leaving the DAO in a permanent decision market state.

**Fix:** `initialize_proposal` and `launch_proposal` now verify that `squads_proposal.transaction_index > squads_multisig.stale_transaction_index`. This prevents proposals from being created or launched if their underlying Squads transaction is already stale.

Also added the same Spot state requirement to `execute_spending_limit_change` for consistency.

### DIFF
--- a/programs/futarchy/src/instructions/admin_approve_execute_multisig_proposal.rs
+++ b/programs/futarchy/src/instructions/admin_approve_execute_multisig_proposal.rs
@@ -19,11 +19,33 @@ pub struct AdminApproveExecuteMultisigProposal<'info> {
     #[account(mut, seeds = [squads_multisig_program::SEED_PREFIX, squads_multisig_program::SEED_MULTISIG, dao.key().as_ref()], bump, seeds::program = squads_multisig_program)]
     pub squads_multisig: Account<'info, squads_multisig_program::Multisig>,
     /// CHECK: squads proposal, initialized by squads multisig program, checked by squads multisig program
-    #[account(mut)]
-    pub squads_multisig_proposal: UncheckedAccount<'info>,
+    #[account(
+        mut,
+        seeds = [
+            squads_multisig_program::SEED_PREFIX,
+            squads_multisig.key().as_ref(),
+            squads_multisig_program::SEED_TRANSACTION,
+            squads_multisig_vault_transaction.index.to_le_bytes().as_ref(),
+            squads_multisig_program::SEED_PROPOSAL,
+        ],
+        bump,
+        seeds::program = squads_multisig_program
+    )]
+    pub squads_multisig_proposal: Account<'info, squads_multisig_program::Proposal>,
     /// CHECK: squads vault transaction, initialized by squads multisig program, checked by squads multisig program
-    #[account(mut)]
-    pub squads_multisig_vault_transaction: UncheckedAccount<'info>,
+    #[account(
+        mut,
+        seeds = [
+            squads_multisig_program::SEED_PREFIX,
+            squads_multisig.key().as_ref(),
+            squads_multisig_program::SEED_TRANSACTION,
+            squads_multisig_vault_transaction.index.to_le_bytes().as_ref(),
+        ],
+        bump,
+        seeds::program = squads_multisig_program
+    )]
+    pub squads_multisig_vault_transaction:
+        Account<'info, squads_multisig_program::VaultTransaction>,
 
     pub squads_multisig_program:
         Program<'info, squads_multisig_program::program::SquadsMultisigProgram>,
--- a/programs/futarchy/src/instructions/admin_approve_execute_multisig_proposal.rs
+++ b/programs/futarchy/src/instructions/admin_approve_execute_multisig_proposal.rs
@@ -60,11 +60,8 @@ impl<'info, 'c: 'info> AdminApproveExecuteMultisigProposal<'info> {
             FutarchyError::InvalidAdmin
         );
 
-        match self.dao.amm.state {
-            PoolState::Spot { .. } => {}
-            _ => {
-                return Err(FutarchyError::PoolNotInSpotState.into());
-            }
+        if !matches!(self.dao.amm.state, PoolState::Spot { .. }) {
+            return Err(FutarchyError::PoolNotInSpotState.into());
         }
 
         Ok(())
--- a/programs/futarchy/src/instructions/admin_approve_execute_multisig_proposal.rs
+++ b/programs/futarchy/src/instructions/admin_approve_execute_multisig_proposal.rs
@@ -38,6 +38,13 @@ impl<'info, 'c: 'info> AdminApproveExecuteMultisigProposal<'info> {
             FutarchyError::InvalidAdmin
         );
 
+        match self.dao.amm.state {
+            PoolState::Spot { .. } => {}
+            _ => {
+                return Err(FutarchyError::PoolNotInSpotState.into());
+            }
+        }
+
         Ok(())
     }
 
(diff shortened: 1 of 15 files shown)


programs/futarchy/src/instructions/admin_approve_execute_multisig_proposal.rs
```diff
@@ -0,0 +1,92 @@
...
+
+    /// CHECK: checked by autocrat program
+    #[account(mut, seeds = [squads_multisig_program::SEED_PREFIX, squads_multisig_program::SEED_MULTISIG, dao.key().as_ref()], bump, seeds::program = squads_multisig_program)]
+    pub squads_multisig: Account<'info, squads_multisig_program::Multisig>,
+    /// CHECK: squads proposal, initialized by squads multisig program, checked by squads multisig program
+    #[account(mut)]
+    pub squads_multisig_proposal: UncheckedAccount<'info>,
+    /// CHECK: squads vault transaction, initialized by squads multisig program, checked by squads multisig program
+    #[account(mut)]
+    pub squads_multisig_vault_transaction: UncheckedAccount<'info>,
```
metaproph3t: why'd you decide not to run the checks on these? all else equal I would prefer it, especially if it's a CU thing

programs/futarchy/src/instructions/admin_approve_execute_multisig_proposal.rs
```diff
@@ -0,0 +1,92 @@
...
+            squads_multisig_vault_transaction,
+            squads_multisig_program,
+            event_authority: _,
+            program: _,
+        } = ctx.accounts;
+
+        let dao_nonce = &dao.nonce.to_le_bytes();
+        let dao_creator_key = &dao.dao_creator.as_ref();
+        let dao_seeds = &[b"dao".as_ref(), dao_creator_key, dao_nonce, &[dao.pda_bump]];
+        let dao_signer = &[&dao_seeds[..]];
```
metaproph3t, line 85: what if we checked here that the dao is in spot state if it's a config transaction?

pileks: I did that, but forgot to push... :)

programs/futarchy/src/instructions/admin_approve_execute_multisig_proposal.rs
```diff
@@ -0,0 +1,99 @@
...
+
+        match self.dao.amm.state {
+            PoolState::Spot { .. } => {}
+            _ => {
+                return Err(FutarchyError::PoolNotInSpotState.into());
+            }
+        }
+
+        Ok(())
+    }
```
pileks: @metaproph3t here is the change related to your comment here:
https://github.com/metaDAOproject/programs/pull/398/files#r2695401559

programs/futarchy/src/instructions/admin_approve_execute_multisig_proposal.rs
pileks: I decided to not run checks, as we don't run them anywhere else.
I've added the checks now!

programs/futarchy/src/instructions/admin_approve_execute_multisig_proposal.rs
```diff
@@ -0,0 +1,121 @@
...
+            metadao_multisig_vault::ID,
+            FutarchyError::InvalidAdmin
+        );
+
+        match self.dao.amm.state {
+            PoolState::Spot { .. } => {}
+            _ => {
+                return Err(FutarchyError::PoolNotInSpotState.into());
+            }
+        }
```
metaproph3t: oh wow this is an annoying rust thing I guess, checking chatgpt to see if it can come up with something cleaner. I know you can't just do the if statement because it's a variable struct thing

metaproph3t:
```
if !matches!(self.dao.amm.state, PoolState::Spot { .. }) {
    return Err(FutarchyError::PoolNotInSpotState.into());
}
```

programs/futarchy/src/instructions/execute_spending_limit_change.rs
```diff
@@ -25,6 +25,13 @@ impl<'info, 'c: 'info> ExecuteSpendingLimitChange<'info> {
     pub fn validate(&self) -> Result<()> {
         require_eq!(self.proposal.state, ProposalState::Passed);
 
+        match self.dao.amm.state {
+            PoolState::Spot { .. } => {}
+            _ => {
+                return Err(FutarchyError::PoolNotInSpotState.into());
+            }
+        }
```
metaproph3t: same thing here

programs/futarchy/src/instructions/initialize_proposal.rs
```diff
@@ -52,6 +53,12 @@ impl InitializeProposal<'_> {
             }
         }
 
+        // Ensure the squads proposal is not invalidated by a previous config transaction
+        require_gt!(
```
metaproph3t, line 57: are we sure this isn't a `require_gte`?

metaproph3t: good to add this check btw

programs/futarchy/src/instructions/launch_proposal.rs
```diff
@@ -27,6 +27,8 @@ pub struct LaunchProposal<'info> {
     pub amm_fail_base_vault: Box<Account<'info, TokenAccount>>,
     #[account(init_if_needed, payer = payer, associated_token::mint = fail_quote_mint, associated_token::authority = dao)]
     pub amm_fail_quote_vault: Box<Account<'info, TokenAccount>>,
+    pub squads_multisig: Account<'info, squads_multisig_program::Multisig>,
+    pub squads_proposal: Account<'info, squads_multisig_program::Proposal>,
```
metaproph3t, line 33: I like the address checks here too

metaproph3t: like I would like those

programs/futarchy/src/instructions/initialize_proposal.rs
pileks: yes, because the invalidation applies by using the config transaction's index:
https://github.com/Squads-Protocol/v4/blob/main/programs/squads_multisig_program/src/state/multisig.rs#L200-L202

This means that the next transaction's index has to be greater than what's written down there.

programs/futarchy/src/instructions/launch_proposal.rs
pileks: I'll add a full check for `squads_multisig`, but I can only add a check for the owner of the proposal to be the Squads Multisig program, because we have no way of knowing what the transaction index is here.

The Squads program performs checks anyways, so it shouldn't be a big deal, but I agree that we can front-load all of the checks we can.

PR #407, opened 2026-01-23
Mint Governor Program

pileks:
This PR introduces a new Solana program that enables delegated minting authority with per-minter limits. Instead of a single wallet holding mint authority, the `mint_governor` program acts as a PDA-based intermediary that can authorize multiple minters, each with an optional cap on how many tokens they can mint.

## Motivation

Projects often need to grant minting capabilities to multiple parties (e.g., DAO proposals, performance package program, etc.) while maintaining central oversight. Directly transferring mint authority is all-or-nothing and irrevocable. This program provides a controlled delegation layer where:

- An admin can grant/revoke minting rights to any address
- Each minter can be capped to a maximum total they can ever mint
- The admin retains the ability to reclaim the underlying mint authority

### Accounts

**MintGovernor** (PDA: `["mint_governor", mint, create_key]`)
- `mint`: The SPL token this governor controls
- `admin`: Address authorized to manage the governor
- `create_key`: Signer used at initialization (allows multiple governors per mint)
- `seq_num`: Monotonic counter for indexing
- `bump`: PDA bump seed

**MintAuthority** (PDA: `["mint_authority", mint_governor, authorized_minter]`)
- `mint_governor`: Parent governor
- `authorized_minter`: Address permitted to call `mint_tokens`
- `max_total`: Optional cap on cumulative minting (None = unlimited)
- `total_minted`: Running total of tokens minted through this authority
- `bump`: PDA bump seed

### Instructions

| Instruction | Signer | Description |
|-------------|--------|-------------|
| `initialize_mint_governor` | `create_key` | Creates the governor account for a mint |
| `transfer_authority_to_governor` | `current_authority` | Transfers SPL mint authority to the governor PDA |
| `add_mint_authority` | `admin` | Creates a MintAuthority for an authorized minter |
| `mint_tokens` | `authorized_minter` | Mints tokens to a destination, respecting limits |
| `update_mint_authority` | `admin` | Modifies `max_total` for an existing authority |
| `remove_mint_authority` | `admin` | Closes a MintAuthority account, reclaiming rent |
| `update_mint_governor_admin` | `admin` | Transfers admin role to a new address |
| `reclaim_authority` | `admin` | Transfers mint authority from governor to any address |

### Minting Flow

1. Admin calls `add_mint_authority` with `max_total = Some(1_000_000)` for address X
2. X calls `mint_tokens` with `amount = 500_000` → succeeds, `total_minted` becomes 500,000
3. X calls `mint_tokens` with `amount = 600_000` → fails with `MintLimitExceeded`
4. X calls `mint_tokens` with `amount = 500_000` → succeeds, `total_minted` becomes 1,000,000
5. Further minting by X is blocked until admin calls `update_mint_authority` to raise the cap

If `max_total = None`, the minter has no cap (though `total_minted` is still tracked for accounting).


programs/mint_governor/src/instructions/initialize_mint_governor.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    #[account(mut)]
+    pub payer: Signer<'info>,
+
+    pub system_program: Program<'info, System>,
+}
+
+impl InitializeMintGovernor<'_> {
+    pub fn validate(&self) -> Result<()> {
+        Ok(())
```
mashedmonday, line 32: So, anyone can create a mint governor?

programs/mint_governor/src/instructions/mint_tokens.rs
```diff
@@ -0,0 +1,113 @@
...
+    pub token_program: Program<'info, Token>,
+}
+
+impl MintTokens<'_> {
+    pub fn validate(&self, args: &MintTokensArgs) -> Result<()> {
+        // Check mint limit if max_total is set
+        if let Some(max_total) = self.mint_authority.max_total {
+            require_gte!(
+                max_total,
+                self.mint_authority.total_minted + args.amount,
```
mashedmonday, line 46: Might want to run this with safe_add instead

                    self.mint_authority.total_minted.safe_add(args.amount),
                    
https://docs.rs/anchor-safe-math/latest/anchor_safe_math/

Both are u64, need to check for overflow.

programs/mint_governor/src/instructions/mint_tokens.rs
```diff
@@ -0,0 +1,113 @@
...
+                    to: ctx.accounts.destination_ata.to_account_info(),
+                    authority: mint_governor.to_account_info(),
+                },
+                signer_seeds,
+            ),
+            args.amount,
+        )?;
+
+        // Update total_minted
+        mint_authority.total_minted += args.amount;
```
mashedmonday, line 85: in case of safe_add, we don't need it again over here, the addition would then be safe.

programs/mint_governor/src/instructions/mint_tokens.rs
```diff
@@ -0,0 +1,113 @@
...
+
+        // Increment seq_num
+        mint_governor.seq_num += 1;
```
mashedmonday, line 88: technically, should use safe_add here as well, but u64 is so big and an overflow, to my current understanding wouldn't break anything, might as well stick with `+`

programs/mint_governor/src/instructions/reclaim_authority.rs
```diff
@@ -0,0 +1,84 @@
...
+        mut,
+        address = mint_governor.mint @ MintGovernorError::MintMismatch
+    )]
+    pub mint: Account<'info, Mint>,
+
+    #[account(address = mint_governor.admin @ MintGovernorError::UnauthorizedAdmin)]
+    pub admin: Signer<'info>,
+
+    /// CHECK: This is the new authority address, no validation needed
+    pub new_authority: UncheckedAccount<'info>,
```
mashedmonday, line 24: Consider checking thew new_authorities SOL balance to confirm it *exists* at least.

programs/mint_governor/src/instructions/transfer_authority_to_governor.rs
```diff
@@ -0,0 +1,62 @@
...
+use crate::{CommonFields, MintAuthorityTransferredEvent, MintGovernor, MintGovernorError};
+
+#[derive(Accounts)]
+pub struct TransferAuthorityToGovernor<'info> {
+    #[account(mut)]
+    pub mint_governor: Account<'info, MintGovernor>,
+
+    #[account(
+        mut,
+        address = mint_governor.mint @ MintGovernorError::MintMismatch
```
mashedmonday, line 14: might want to add a constraint to ensure the mint's authority == current_authority and throw a proper Error, otherwise you get an error from CPI

programs/mint_governor/src/instructions/mint_tokens.rs
pileks: Thanks for the comment! Anchor already does this thanks to the `overflow-checks` release flag in `Cargo.toml`, so we're safe there.

programs/mint_governor/src/instructions/initialize_mint_governor.rs
pileks: Technically, yes. No reason not to, however I don't know why they would do so.
Frontrunning is protected by the `create_key` which needs to be a signer on mint governor creation.

programs/mint_governor/src/instructions/reclaim_authority.rs
pileks: This would prevent us from assigning authority to a Squads multisig vault though, as that one is just a derived PDA without any balance - just used for signing.

xeroc: good point.

programs/mint_governor/src/instructions/mint_tokens.rs
xeroc: TIL ..

PR #409, opened 2026-01-27
Increase timeout for awaiting setup in anchor test

pileks:
...Tests were intermittently failing with awaits on slower machines, I just gave it a bit more time to set up.


pileks:
Decided against merging for now. If tests appear to be failing for whatever reason in the future will reopen.

PR #415, opened 2026-02-02
Remove proposal

pileks:


programs/futarchy/src/instructions/admin_remove_proposal.rs
```diff
@@ -0,0 +1,50 @@
...
+    pub dao: Box<Account<'info, Dao>>,
+    #[account(mut)]
+    pub admin: Signer<'info>,
+}
+
+impl AdminRemoveProposal<'_> {
+    pub fn validate(&self) -> Result<()> {
+        #[cfg(feature = "production")]
+        require_keys_eq!(self.admin.key(), admin::ID, FutarchyError::InvalidAdmin);
+
```
metaproph3t, line 23: this is fine for now, but we should add a comment that maybe we want to extend this in the future for non-draft proposals

R-K-H (approved):

PR #425, opened 2026-02-13
fix: remove cfg(production) admin bypass + merge pass pool reserves o…

0xksure:
…n cancel

Two HIGH severity security fixes:

1. Admin functions bypass in non-production mode (FT-F02) All admin functions (admin_remove_proposal, admin_cancel_proposal, admin_approve_execute_multisig_proposal, collect_fees) gate admin key verification behind #[cfg(feature = "production")]. If deployed without the production feature flag, ANY signer can execute admin operations including cancelling active proposals and collecting protocol fees. Fix: Always check admin keys at runtime.

2. admin_cancel_proposal drops pass pool reserves (FT-F09) When admin cancels a proposal, only fail pool reserves are merged back to spot. Pass pool reserves are silently dropped via Rust's '..' pattern destructure, permanently stranding tokens in the AMM vault. Fix: Also merge pass pool reserves back to spot accounting.

More information https://github.com/0xksure/solana-audit-report/blob/master/PR_BODY_FUTARCHY.md

### DIFF
--- a/SECURITY_AUDIT.md
+++ b/SECURITY_AUDIT.md
@@ -1,0 +1,199 @@
+## Security Audit: 2 HIGH Severity Findings in Futarchy Program
+
+### Summary
+
+Independent security audit of the futarchy program identified **2 HIGH severity vulnerabilities** that could lead to unauthorized admin access and permanent fund loss.
+
+---
+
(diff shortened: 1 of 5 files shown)


R-K-H:
1. Isn't an issue given only production should include administrative fees, others deploying shouldn't utilize it.

2. This isn't how conditional markets work.

PR #430, opened 2026-02-21
project-wide claude permissions (read actions)

pileks:


R-K-H (approved):
If you fix the lock, then we're good.

PR #446, opened 2026-03-30
refactor(futarchy): replace vault transaction account with transaction_index arg

pileks:
## Summary

Replaces the `vault_transaction` account in `admin_approve_multisig_proposal` with a `transaction_index` argument, enabling approval of Squads multisig proposals via batch transactions.

## Changes

- **`admin_approve_multisig_proposal.rs`**: Removed the `vault_transaction` account from the `AdminApproveMultisigProposal` accounts struct. Instead, the instruction now accepts `AdminApproveMultisigProposalArgs` containing a `transaction_index: u64`. The `squads_multisig_proposal` PDA is now derived using this index directly, rather than reading it from a passed-in vault transaction account.
- **`lib.rs`**: Updated the instruction signature to accept and forward the new `args` parameter, with `access_control` validation wired through `validate(&args)`.
- **SDK & Tests**: Regenerated SDK types and updated tests to pass `transactionIndex` instead of the `vaultTransaction` account.

## Motivation

Previously, approving a multisig proposal required passing the full `vault_transaction` account, which prevented batching multiple approve calls in a single Solana transaction (since each would reference a different vault transaction account that needed to be resolved). By taking the transaction index as a simple `u64` argument and deriving the proposal PDA on-chain, callers can now batch multiple approvals without needing to supply additional accounts.


pileks:
Despite us likely not needing it immediately, I still think we should add this. It is a minor change that allows us to run batch txns when we need them.

cc @R-K-H lmk if you agree

PR #463, opened 2026-05-28
Note deploy Claude skill

metapileks:
Adds a `/deploy-notes` skill to Claude that writes a neat message based on our program deploy workflows and proposals.

- `/deploy-notes <program>` — **preview**: what's about to ship since the last successful mainnet deploy of `<program>`.
- `/deploy-notes <program> --compare [<old-run-id> <new-run-id>]` — **compare**: write release notes between two past deploys (default: the last two successful runs). Handles the first-deploy case automatically.

You can also add a `--md` flag to print out a markdown doc with these notes.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -44,7 +44,8 @@
       "Bash(gh pr list:*)",
       "Bash(gh pr view:*)",
       "Bash(gh pr diff:*)",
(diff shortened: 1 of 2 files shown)


metanallok (approved):
No notes this is perfect

PR #470, opened 2026-06-29
added laso and some helpers to improve xp

metanallok:
Awaiting confirmation from Laso, but otherwise the primary pieces are here.

- [ ] Need to write approval script for Laso, expect it to approve fixed amount for a few addresses then pro-rata for the rest. We could experiment with pro-rata time based accumulator or something, but that's offchain.


metapileks (approved):
LGTM

Nit: We have this repeating pattern of simulating then sending off the tx with simulation units + 20%. Maybe we can refactor that too.

metanallok:
Updating the ledger stuff beyond this package version is only 3 days old, so would rather not. But can resolve in short order. I agree on the common features / libs since we've been using it a lot now, there's a question if it belongs in the SDK in my mind given the fact that we need it for scripts and likely just want coverage on many functions.

We'll see how it goes but since I've been using it a lot I think I've got some ideas.

metapileks:
> Updating the ledger stuff beyond this package version is only 3 days old, so would rather not. But can resolve in short order. I agree on the common features / libs since we've been using it a lot now, there's a question if it belongs in the SDK in my mind given the fact that we need it for scripts and likely just want coverage on many functions.
> 
> We'll see how it goes but since I've been using it a lot I think I've got some ideas.

My hesitance with adding it to the SDK as some form of helper collection is that we're then implying it inside our SDK's semver. And these are things that teams usually ship for themselves. But we can get you set up with a separate collection of scripts if you want these. 😄

metapileks (approved):

PR #475, opened 2026-07-19
fix: improve broken devcontainer

0xahzam:
- existing devcontainer doesn't build without errors
- fixed installs, added devcontainer.json and updated readme
- anchor build passes inside the devcontainer now
- should be helpful for future contributors


metapileks:
We currently don't accept outside contributions outside of special circumstances. You are more than welcome to apply these changes to your own fork if you want to experiment with the protocol.

PR #477, opened 2026-07-22
Mini-instructions (MVP)

metapileks:
Replaces futarchy v0.6's one-size-fits-all proposals with a fixed catalog of typed governance actions. For every type except the `execute_arbitrary` catch-all, futarchy builds the Squads vault transaction itself from a per-type template at create.

## The catalog

| Action | Who can propose | Market | TWAP start delay | Threshold | Council can block | Cooldown |
|---|---|---|---|---|---|---|
| `large_spend(amount)` | Required to launch | 1.5 days | 12 h | −10% | Yes | — |
| `mint_tokens(amount, recipient)` | - | 5 days | 24 h | +5% | Yes | — |
| `spending_limit_change(config)` | Required to launch | 5 days | 24 h | +5% | Yes | — |
| `execute_arbitrary` | - | 10 days | 24 h | +10% | Yes | — |
| `hostile_takeover(new_team, limit_action)` | - | 20 days | 24 h | +10% | No | 20 days after failure |
| `hostile_liquidate(liquidator)` | - | 10 days | 24 h | +25% | No | 10 days after failure |
| `buyback_token(amount, schedule, price_band)` | - | 10 days | 24 h | +10% | Yes | 90 days after any finalization |

## Program changes

**Added**
- `initialize_*_proposal` — permissionless creates that build the vault tx from the type's template (`typed_initialize.rs`).
- `initialize_buyback_token_proposal` — the buyback "template" is a `metadao-buyback/1` memo recording the DCA schedule for off-chain execution against Jupiter's Trigger API; create validates the shape (a whole number of cycles, at least two; cycle between 1 minute and 1 year; start delay ≤ 30 days; optional min/max price band)
- `set_spending_limit` — vault-signed writer of the on-DAO spending-limit record (replaces `execute_spending_limit_change`, which bypassed the record)
- `sync_spending_limit` — permissionless dirty-gated crank projecting the record onto the Squads `SpendingLimit` (futarchy signs as config authority)
- `apply_liquidation` — vault-signed and kind-checked (requires a passed `HostileLiquidate` proposal, so liquidation can't ride `execute_arbitrary`): installs the liquidator, zeroes the limit record, sweeps the treasury's own AMM position; the vault tx pairs it with a memo recording the legal-layer IP transfer back to the original team
- `admin_update_proposal_params` — ops-multisig-signed retune of a single **draft** `execute_arbitrary` proposal's `duration_in_seconds` and `pass_threshold_bps`.
- State: typed `ProposalAction` on `Proposal` (borsh tag doubles as the kind) + snapshotted signed threshold and blockable flag; `Dao` gains `liquidator: Option<Pubkey>`, per-hostile last-failed and last-buyback-finalized timestamps, `spending_limit_dirty`, and `initial_spending_limit` promoted to the authoritative spending-limit record

**Modified**
- `initialize_proposal` is now the `execute_arbitrary` path (externally supplied, unvalidated, at the stricter uniform 10d/+10%)
- `launch_proposal` — kind-aware sponsorship and cooldown checks alongside the DAO's existing anti-grief gate; writes the kind's TWAP start delay into both conditional oracles, and rejects a market that isn't strictly longer than its own delay (`ProposalDurationTooShort`); runs the kind's launch-account gate over remaining accounts
- `finalize_proposal` — uses the proposal's snapshotted threshold (not DAO config), stamps hostile failure timestamps, and stamps the buyback timestamp on either outcome (the 90-day cooldown rate-limits treasury draws the DAO consented to, not failed retries; a council cancel stamps nothing)
- `admin_cancel_proposal` — respects the blockable flag (council can cancel routine types mid-market, never hostile ones)
- `update_dao` — spending-limit-neutral; `team_address` only gates sponsorship
- Liquidated guards on state-mutating instructions (third-party `withdraw_liquidity` stays open); liquidated DAOs execute via a liquidator-gated enqueue → permissionless approve → ordinary Squads execution
- Per-DAO `pass_threshold_bps` / `seconds_per_proposal` / `team_sponsored_pass_threshold_bps` / `twap_start_delay_seconds` become vestigial.

**Removed**
- Optimistic proposal machinery: `initiate_vault_spend_optimistic_proposal`, `finalize_optimistic_proposal`, and their special cases in create/launch/admin paths
- `execute_spending_limit_change`

## Migration

- `resize_dao` / `resize_proposal` cranks; legacy proposals snapshot to `ExecuteArbitrary` from the vestigial per-DAO fields, optimistic values cleared

## SDK

- Crate 0.6.1 → 0.6.2; previous IDL frozen as `v0.6.0`, new `v0.6.1` IDL snapshot added alongside it
- `FutarchyClient` methods and PDA helpers for every new instruction

<h3>Greptile Summary</h3>

Introduces typed futarchy governance actions and their fixed market parameters, replacing the previous one-size-fits-all proposal flow.
- Builds action-specific Squads vault transactions during proposal creation.
- Adds authoritative spending-limit state with permissionless synchronization to Squads.
- Adds hostile takeover and liquidation flows, including terminal liquidated-DAO guards and liquidator-controlled execution.
- Adds migration cranks, SDK support, IDL snapshots, and extensive unit and integration coverage.

<h3>Confidence Score: 5/5</h3>

The PR appears safe to merge based on the reviewed typed-action bindings, migration behavior, liquidation lifecycle, and spending-limit synchronization.

The changed proposal templates bind action payloads and target accounts into their Squads transactions, liquidation is constrained to passed hostile-liquidation proposals with terminal replay protection, and spending-limit synchronization preserves atomic state through Solana transaction rollback.

<details><summary><h3>Important Files Changed</h3></summary>

| Filename | Overview |
|----------|----------|
| programs/futarchy/src/instructions/typed_create.rs | Centralizes permissionless typed proposal creation and binds generated Squads transactions to proposal action snapshots. |
| programs/futarchy/src/instructions/apply_liquidation.rs | Applies passed hostile liquidation proposals, installs the liquidator, disables spending rights, and sweeps the treasury-owned AMM position. |
| programs/futarchy/src/instructions/sync_spending_limit.rs | Dirty-gates projection of the authoritative DAO spending-limit record onto the corresponding Squads account. |
| programs/futarchy/src/instructions/launch_proposal.rs | Adds action-specific sponsorship and hostile cooldown checks while retaining liquidity and Squads-state validation. |
| programs/futarchy/src/instructions/finalize_proposal.rs | Finalizes markets using snapshotted thresholds and records failed hostile-action cooldown timestamps. |
| programs/futarchy/src/instructions/resize_proposal.rs | Migrates legacy proposals with their existing threshold semantics and classifies them as blockable arbitrary-execution proposals. |
| programs/futarchy/src/state/proposal_action.rs | Defines the typed governance action catalog and fixed action-specific governance parameters. |
| sdk/src/futarchy/v0.6/FutarchyClient.ts | Adds SDK builders and account derivations for typed proposal, spending-limit, liquidation, and migration operations. |

</details>

<sub>Reviews (1): Last reviewed commit: ["update idl, reduce cycles in tests"](https://github.com/metadaoproject/programs/commit/a298864cfb5686a2fca70a40ec00685337910aec) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=46843100)</sub>


programs/futarchy/src/instructions/admin_cancel_proposal.rs
```diff
@@ -69,6 +69,14 @@ pub struct AdminCancelProposal<'info> {
...
 impl AdminCancelProposal<'_> {
     pub fn validate(&self) -> Result<()> {
+        // Unblockable proposals are censorship-proof once live: nobody, including
+        // the council, can cancel them. Reads the create-time snapshot so a
+        // live proposal keeps the flag it launched with.
+        require!(
+            self.proposal.council_can_block,
+            FutarchyError::InvalidProposalKind
+        );
+
```
metaproph3t, line 79: I feel like admin should be more for security stuff and we can have a different one for council blocking proposals, maybe? Not a strong opinion

programs/futarchy/src/instructions/admin_enqueue_multisig_proposal_approval.rs
```diff
@@ -62,17 +62,28 @@ pub struct AdminEnqueueMultisigProposalApproval<'info> {
 
 impl AdminEnqueueMultisigProposalApproval<'_> {
     pub fn validate(&self, _args: &AdminEnqueueMultisigProposalApprovalArgs) -> Result<()> {
-        #[cfg(feature = "production")]
-        require_keys_eq!(self.admin.key(), admin::ID, FutarchyError::InvalidAdmin);
+        // On a liquidated DAO the liquidator replaces the admin id as the
+        // required signer. Enqueueing is the only capability the liquidator
+        // gains: the approve leg stays permissionless and execution is
+        // ordinary top-level Squads execution.
```
metaproph3t, line 68: How does this work?

programs/futarchy/src/instructions/apply_liquidation.rs
```diff
@@ -0,0 +1,197 @@
...
+                }
+
+                dao.amm.total_liquidity -= liquidity_to_sweep;
+
+                position.liquidity = 0;
+                {
+                    let mut data = amm_position.try_borrow_mut_data()?;
+                    let mut writer: &mut [u8] = &mut data;
+                    position.try_serialize(&mut writer)?;
+                }
```
metaproph3t: Feels janky

programs/futarchy/src/instructions/apply_liquidation.rs
```diff
@@ -0,0 +1,197 @@
...
+                crate::ID,
+                anchor_lang::error::ErrorCode::AccountOwnedByWrongProgram
+            );
+
+            let mut position: AmmPosition = {
+                let data = amm_position.try_borrow_data()?;
+                AmmPosition::try_deserialize(&mut &data[..])?
+            };
+
+            if position.liquidity > 0 {
```
metaproph3t, line 125: Do we not want to internally call the withdraw liquidity instruction?

programs/futarchy/src/instructions/finalize_proposal.rs
```diff
@@ -164,6 +160,18 @@ impl FinalizeProposal<'_> {
 
         proposal.state = new_proposal_state;
 
+        if new_proposal_state == ProposalState::Failed {
+            match proposal.action {
+                ProposalAction::HostileTakeover { .. } => {
+                    dao.last_failed_takeover_at = clock.unix_timestamp
+                }
+                ProposalAction::HostileLiquidate { .. } => {
+                    dao.last_failed_liquidation_at = clock.unix_timestamp
```
metaproph3t, line 171: What's the point of these?

programs/futarchy/src/instructions/initialize_hostile_liquidate_proposal.rs
```diff
@@ -0,0 +1,81 @@
...
+    pub create: TypedCreateAccounts<'info>,
+}
+
+impl InitializeHostileLiquidateProposal<'_> {
+    pub fn validate(&self) -> Result<()> {
+        self.create.validate()
+    }
+
+    pub fn handle(ctx: Context<Self>, args: InitializeHostileLiquidateProposalArgs) -> Result<()> {
+        let create = &mut ctx.accounts.create;
```
metaproph3t: what is this object?

metaproph3t: feels a little smelly

metaproph3t: May just need a better name

programs/futarchy/src/instructions/initialize_hostile_liquidate_proposal.rs
```diff
@@ -0,0 +1,81 @@
...
+                ),
+                token_program: token::ID,
+                event_authority,
+                program: crate::ID,
+            }
+            .to_account_metas(None),
+            data: crate::instruction::ApplyLiquidation.data(),
+        };
+
+        let event = create.create_proposal(
```
metaproph3t: We *may* want to add a memo instruction that's like *transfer IP back to team*

programs/futarchy/src/instructions/initialize_large_spend_proposal.rs
```diff
@@ -0,0 +1,67 @@
...
+    pub amount: u64,
+}
+
+#[derive(Accounts)]
+#[event_cpi]
+pub struct InitializeLargeSpendProposal<'info> {
+    pub create: TypedCreateAccounts<'info>,
+}
+
+impl InitializeLargeSpendProposal<'_> {
```
metaproph3t, line 14: I think this is one that the team should only be able to create

programs/futarchy/src/instructions/initialize_mint_tokens_proposal.rs
```diff
@@ -0,0 +1,129 @@
...
+    #[account(address = create.dao.base_mint)]
+    pub base_mint: Box<Account<'info, Mint>>,
+    /// Only for governed mints (v0.8 launches): the `MintGovernor` holding the
+    /// base mint's authority.
+    pub mint_governor: Option<Box<Account<'info, mint_governor::MintGovernor>>>,
+    /// Only for governed mints: the vault's minting rights on `mint_governor`.
+    pub mint_authority: Option<Box<Account<'info, mint_governor::MintAuthority>>>,
+}
+
+impl InitializeMintTokensProposal<'_> {
```
metaproph3t, line 26: Same here, I think this should be only team

programs/futarchy/src/instructions/initialize_mint_tokens_proposal.rs
```diff
@@ -0,0 +1,129 @@
...
+#[event_cpi]
+pub struct InitializeMintTokensProposal<'info> {
+    pub create: TypedCreateAccounts<'info>,
+    #[account(address = create.dao.base_mint)]
+    pub base_mint: Box<Account<'info, Mint>>,
+    /// Only for governed mints (v0.8 launches): the `MintGovernor` holding the
+    /// base mint's authority.
+    pub mint_governor: Option<Box<Account<'info, mint_governor::MintGovernor>>>,
+    /// Only for governed mints: the vault's minting rights on `mint_governor`.
+    pub mint_authority: Option<Box<Account<'info, mint_governor::MintAuthority>>>,
```
metaproph3t, line 23: Smart

metaproph3t:
Initial pass completed

I think we should have some inline docs at this point in lib.rs to explain everything

programs/futarchy/src/instructions/finalize_proposal.rs
metapileks: We want to prevent hostile proposals from being retried immediately, as they are "out of our control" in terms of us vetoing them.

metapileks: Added a comment describing this.

programs/futarchy/src/instructions/initialize_hostile_liquidate_proposal.rs
metapileks: Renamed to `TypedInitializeAccounts` and `typed_initialize_accounts` respectively, as it will be clearer that we're talking about a typed initialization and its associated common accounts, as opposed to a "generic" initialization. Didn't like `common` for this, as it's not descriptive enough.

programs/futarchy/src/instructions/initialize_hostile_liquidate_proposal.rs
metapileks: Agreed, will add.

programs/futarchy/src/instructions/admin_cancel_proposal.rs
metapileks: I agree. Mainly holding this here until we make the decision on whether we're merging in #469 so that we know whether that will be the council (one of the parties in the multisig), or we will have a separate council that can cancel a proposal at any time.

programs/futarchy/src/instructions/apply_liquidation.rs
metapileks: First reason is that Anchor's serialization approach, by design, doesn't catch changes that happen in inner instructions. We have to manually call `.reload()` on all accounts that we need to touch after the CPI, and any accounts we touch before the CPI would not properly reflect within it, as their changes are not yet written to account data.

Second reason is that we'd see two events - one "regular" `WithdrawLiquidityEvent` and later another `ApplyLiquidationEvent`, instead of just the `ApplyLiquidationEvent`.

I've now changed it so that it uses a `withdraw_from_position` function shared between `withdraw_liquidity` and `apply_liquidation`, whose purpose is to run the withdrawal logic (including token transfers). This way we still emit a single event per instruction, and we contain (most) of the account updates. The only exception is the `UncheckedAccount` for the position in `apply_liquidation`, but that is a simple addition and now it's properly documented.

programs/futarchy/src/instructions/apply_liquidation.rs
metapileks: Let me know if it feels a bit less jank now that there's an explicit comment and less code around it.

programs/futarchy/src/instructions/initialize_large_spend_proposal.rs
metapileks: So you are thinking both create AND launch should be gated? Right now anyone can create these, but only the team can approve launching these, as they need to sponsor them.

programs/futarchy/src/instructions/initialize_mint_tokens_proposal.rs
metapileks: See https://github.com/metaDAOproject/programs/pull/477#discussion_r3717437120

programs/futarchy/src/instructions/admin_enqueue_multisig_proposal_approval.rs
metapileks: We reuse our existing system of enqueueing and executing an admin-controller Squads proposal approval, but now whoever is the liquidator can do the enqueueing of the approval. Running the approval itself is always permissionless.
The reason for it having two instructions (enqueue approval, execute approval) is because that way we can enqueue using another Squads multisig then permissionlessly approve, thus avoiding reentrancy into Squads (if it were a single instruction).
`Squads (liquidator/admin) -> Futarchy -> Squads (DAO)` is the reentrancy path we're avoiding.

